PR #8, opened 2015-04-20
OpenSL ES buffer callbacks on Android

janblome:
On some Android implementations of OpenSL ES the playback buffer callback can be called even after playback was stopped. This can lead to a crash when a callback happens after the buffer was freed. I could observe this behavior using the Android Emulator with API 17 and on a real device running API 19.

The Android NDK docs on OpenSL mention a similar behavior:
> A corollary is that it is unspecified whether buffer queue callbacks are called upon transition to SL_PLAYSTATE_STOPPED or by BufferQueue::Clear. We recommend that you do not rely on either behavior; be prepared to receive a callback in these cases, but also do not depend on receiving one.

To prevent this crash I changed the OpenSL ES back-end to free the buffer only after the buffer queue object was destroyed. I also moved the buffer allocation from the start to the reset function. With these changes I could not observer any further crashes.

### DIFF
--- a/Alc/backends/opensl.c
+++ b/Alc/backends/opensl.c
@@ -280,12 +280,29 @@ static ALCboolean opensl_reset_playback(ALCdevice *Device)
         result = VCALL(data->bufferQueueObject,Realize)(SL_BOOLEAN_FALSE);
         PRINTERR(result, "bufferQueue->Realize");
     }
+    
+    if(SL_RESULT_SUCCESS == result)
+    {
+        free(data->buffer);
+        data->frameSize = FrameSizeFromDevFmt(Device->FmtChans, Device->FmtType);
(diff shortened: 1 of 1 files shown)


kcat:
The changes look okay, but if OpenSL allows the callback to be called after playback was stopped, that's a bigger problem that needs to be fixed. OpenAL Soft assumes its aluMixData function (which is called in the callback to mix audio) won't be called after the stop method returns. In particular, it's going to be changing the device properties, reallocating mixer buffers, calling reset, and calling start without the device lock, so the callback must be finished and must not be invoked anymore once ```opensl_stop_playback``` returns.

janblome:
I see, in that case this is not really a good solution. I created a new pull request (#9) with an alternate solution.

PR #9, opened 2015-04-21
OpenSL ES buffer callbacks on Android

janblome:
This is an alternate solution to the problem described in pull request #8.

I changed the stop function to destroy the buffer queue object before freeing the buffer. Calling `Destroy`  on the buffer queue seems to block until any running callbacks are done and prevent any further callbacks. I also moved the realization of the buffer queue object from the reset to the start function. This way the start function can re-realize the object after it was destroyed by the stop function.

### DIFF
--- a/Alc/backends/opensl.c
+++ b/Alc/backends/opensl.c
@@ -205,10 +205,6 @@ static void opensl_close_playback(ALCdevice *Device)
 {
     osl_data *data = Device->ExtraData;
 
-    if(data->bufferQueueObject != NULL)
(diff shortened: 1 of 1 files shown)


kcat:
Doesn't the Destroy method invalidate the bufferQueueObject? It's not guaranteed that the reset method will be called in between stop and start.

janblome:
Yes, you're right. My mistake. Since in that case I'm not sure how to resolve this problem, i will open an issue instead.

PR #39, opened 2016-04-30
ALc.c: ProbeDevices: Avoid crash when no capture device is available.

fbarriga:
When no capture devices are available running `openal-info` crashes. This patch add an extra check.
This is generating the device that generate the crash:
https://github.com/kcat/openal-soft/blob/master/Alc/ALc.c#L1057-L1064


kcat:
Thanks for reporting this. There's a fix now in commit d1e98c3 (which I decided to do a bit differently, to avoid potential warning spam).

Issue #55, opened 2016-08-08
[Feature Request] AL_MAX_GAIN > 1.0


flibitijibibo:
In XACT, volumes can range from -96.0dB to 6.0dB. This means that the amplitude ratio can range from [0, 2] instead of OpenAL's current limitation of [0,1]. FNA's XACT implementation currently calls alSourcef requesting higher AL_GAIN values, but because of the AL_MAX_GAIN value being clamped, we're stuck with a cap lower than what we need.

I was able to verify this locally by just skipping the clamps altogether...

https://gist.github.com/flibitijibibo/3f46b07df5e0ebba17b25a83d2581b6c

... but I'd love to be able to set AL_MAX_GAIN to higher values! Of course, the spec has stated the limit since 1.0, so I know it'd be a tough change to make (on top of the fact that having a hard maximum is kind of important for safety reasons!). Is there any way we could sidestep this without hacking Alu.c?

kcat:
For OpenAL Soft, a quick'n'dirty workaround would be to double the desired listener gain and halve the desired gain of *all* sources. So a normal listener gain of 1.0 becomes 2.0, and a normal source gain of 1.0 becomes 0.5, which results in a normal mix gain of 1.0. But a desired source gain of 2.0 to become 1.0, which results in a mix gain of 2.0. The source gain clamping happens before the listener gain is applied (after distance attenuation, before filters and listener gain), so the listener gain can raise a source's output above 0dB.

I certainly wouldn't mind adding an extension to allow AL_MAX_GAIN > 1.0, though I wonder if the max limit should be raised to 2, 4, etc, or be completely unbounded.

flibitijibibo:
A possible extension could be something like AL_SOFT_source_gain_clamp_control, where a context could specify both the lowest MIN and the highest MAX, allowing more flexibility with these parameters but still allowing boundaries to prevent dangerous volumes from being played. In our case, because volumes can be multiplied on top of each other, there's definitely a risk of 6dB values constantly multiplying into something really loud, so I'd probably cap it _somewhere_ for safety reasons, but unfortunately I don't know what a fixed value would be. In one instance, for example, the final AL_GAIN for a single XACT cue turned out to be ~5.5! Made sense for this dialog WaveBank, but yeah... pretty scary for a properly normalized metal song. :P

flibitijibibo:
Here's an initial scribbling of such an extension:

```
diff --git a/Alc/ALc.c b/Alc/ALc.c
[...]

PR #59, opened 2016-08-29
AL_SOFT_source_gain_clamp_ex

flibitijibibo:
This is a formal pull request for #55, based on the current extension proposal as of writing:

http://openal.org/pipermail/openal/2016-August/000530.html


kcat:
Thanks, but I already had code to do this. Note though that until the extension is finalized, it won't be public in alext.h, the name will temporarily be ```SOFTX``` instead of ```SOFT```, and there's a potential for a behavior change. In other words, feel free to play around with it, but please wait until it's finalized before using it in production code (which shouldn't be too long).

Issue #54, opened 2016-08-07
armv7l compile issues with arm_neon.h


l3eta:
Make Output:
[alarm@alarm build]$ make
Scanning dependencies of target makehrtf
[  1%] Building C object CMakeFiles/makehrtf.dir/utils/makehrtf.c.o
[  3%] Linking C executable makehrtf
[  3%] Built target makehrtf
Scanning dependencies of target common
[  4%] Building C object CMakeFiles/common.dir/common/almalloc.c.o
[  6%] Building C object CMakeFiles/common.dir/common/atomic.c.o
[  8%] Building C object CMakeFiles/common.dir/common/rwlock.c.o
[  9%] Building C object CMakeFiles/common.dir/common/threads.c.o
[ 11%] Building C object CMakeFiles/common.dir/common/uintmap.c.o
[ 12%] Linking C static library libcommon.a
[ 12%] Built target common
Scanning dependencies of target openal
[ 14%] Building C object CMakeFiles/openal.dir/OpenAL32/alAuxEffectSlot.c.o
/media/sdcard/workspace/openal-soft/OpenAL32/alAuxEffectSlot.c: In function 'RemoveEffectSlotList':
/media/sdcard/workspace/openal-soft/OpenAL32/alAuxEffectSlot.c:431:9: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
     if(!ATOMIC_COMPARE_EXCHANGE_STRONG(ALeffectslot*, &context->ActiveAuxSlotList, &root, next))
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/media/sdcard/workspace/openal-soft/OpenAL32/alAuxEffectSlot.c:437:18: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
         } while(!ATOMIC_COMPARE_EXCHANGE_STRONG(ALeffectslot*, &cur->next, &root, next));
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[ 16%] Building C object CMakeFiles/openal.dir/OpenAL32/alBuffer.c.o
[ 17%] Building C object CMakeFiles/openal.dir/OpenAL32/alEffect.c.o
[ 19%] Building C object CMakeFiles/openal.dir/OpenAL32/alError.c.o
[ 20%] Building C object CMakeFiles/openal.dir/OpenAL32/alExtension.c.o
[...]

PR #64, opened 2016-09-06
Build NEON code with -mfpu=neon

tpetazzoni:
The ARM-specific NEON code needs to be built with -mfpu=neon to avoid
build failures when a difference FPU is used by default by the
compiler.

Fixes issue #54.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>


kcat:
Are there any potential ABI issues by building one source file with ```-mfpu=neon``` and others without? That is, would sources compiled with that switch expect float parameters to be passed differently, or expect the FPU to be in a certain kind of state, compared to sources compiled without that switch?

tpetazzoni:
@kcat I don't *think* there are ABI issues. The floating point ABI is defined by -mfloat-abi (soft, softfp, hard), and the floating point ABI is what defines how floating point arguments are passed between functions. And the linker anyway doesn't allow you to link -mfloat-abi=softfp code with -mfloat-abi=hard code. So I *think* building just one file with -mfpu=neon is correct (of course, assuming the HW actually has NEON support), but I am not a definitive expert on the topic.

kcat:
You have tested it, at least? Made sure the NEON mixers work with this?

tpetazzoni:
@kcat No, I only did build testing. I'm not using OpenAL, I'm just submitting this because we're facing build issues in Buildroot. I've asked the people using OpenAL in Buildroot to test the patch, we'll see what they say.

kcat:
I think I may need to do runtime detection for NEON too, otherwise a lib built with NEON support will crash if run on a device without it (rather than falling back to the C methods). Or simply leave NEON off by default, but then devices that have NEON won't take advantage of it.

tpetazzoni:
VLC relies on /proc/cpuinfo, see http://git.videolan.org/?p=vlc.git;a=blob;f=src/linux/cpu.c;h=38fcce98f8526f2b0dba7bbfd64a36f7e3ecefeb;hb=HEAD#l45

PR #69, opened 2016-09-12
Test for -mfpu=neon before using it

AndreRH:
Otherwise it fails with aarch64 toolchains which have arm_neon.h but don't support -mfpu=neon


AndreRH:
see for example http://autobuild.buildroot.net/results/0e8/0e8815b79887f9abe04af844178c0ed788767d39/build-end.log

kcat:
I was going to ask if you could make it check similar to how ```-msse``` is handled, but since I was working on this code anyway, I fixed it with commit 46b3e1d08ca10e809eb2e20b6371812958b97e1f. Let me know if there's still a problem.

PR #71, opened 2016-09-16
windows: updated dsound search paths

snikulov:
Added more hints for Direct Sound search.
It will help on computers where DirectX SDK not installed, but present in Windows SDK.

### DIFF
--- a/cmake/FindDSound.cmake
+++ b/cmake/FindDSound.cmake
@@ -8,24 +8,30 @@
 #   DSOUND_LIBRARY     - the dsound library
 #
 
+set(_pf86 "ProgramFiles(x86)")
+
 find_path(DSOUND_INCLUDE_DIR
-          NAMES dsound.h
-          PATHS "${DXSDK_DIR}"
-          PATH_SUFFIXES include
-          DOC "The DirectSound include directory"
-)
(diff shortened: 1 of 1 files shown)


kcat:
Having the version explicitly in the path doesn't seem like a good idea. Wouldn't that require changing it for every release of the SDK?

snikulov:
@kcat DirectSound is deprecated, do it's little possibility to be changed in future. From vista it should be WASAPI.

snikulov:
@kcat it's just a hint for CMake. It's just a list of well known paths

kcat:
DirectSound is deprecated, but the Windows SDK can still be updated and quickly make that path out of date. Is it always installed into ```Program Files (x86)``` too, and never ```Program Files``` or somewhere else?

snikulov:
@kcat Then we add another hint. But it'll be unlikely because DS was removed since 7.

kcat:
Pull request #78, with commit d142ba1ab4d5d63f166f9236da21460ec90ce9bd, includes the WindowsSDK paths when searching for DSound. Does that work for you too?

snikulov:
@kcat I've moved from Windows development, so I'm unable to verify it.
Let's close this.

PR #78, opened 2016-10-30
Added cmake support for dsound and windows 8/10 SDKs

septag:
I know dsound may not be used as the first backend, and don't know if it's any use for you.
But I updated cmake file to find dsound on windows
Thanks

### DIFF
--- a/cmake/FindDSound.cmake
+++ b/cmake/FindDSound.cmake
@@ -8,19 +8,63 @@
 #   DSOUND_LIBRARY     - the dsound library
 #
 
+# DSOUND_INCLUDE_DIR
+file(GLOB DXSDK_INCLUDE_DIRS LIST_DIRECTORIES TRUE "C:/Program Files (x86)/Windows Kits/10/Include/*")
+if (DXSDK_INCLUDE_DIRS)
+  list(SORT DXSDK_INCLUDE_DIRS)
+  list(REVERSE DXSDK_INCLUDE_DIRS)
+endif() 
+
 find_path(DSOUND_INCLUDE_DIR
-          NAMES dsound.h
-          PATHS "${DXSDK_DIR}"
-          PATH_SUFFIXES include
-          DOC "The DirectSound include directory"
+          NAMES 
+            "dsound.h" 
+          PATHS 
+            "${DXSDK_DIR}"
+            ${DXSDK_INCLUDE_DIRS}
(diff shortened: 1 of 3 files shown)


kcat:
This looks really odd, that it gets all files and directories in specific directories, then does a reverse sort on the list. The ```um``` path suffix looks weird too. Doesn't the SDK have a batch file or something that sets the DXSDK_DIR environment variable (or set it during installation)? Hard-coding the paths isn't a good idea because it may not be installed to ```C:/Program Files (x86)```, I think there's an environment variable for that, and does the full install path depend on the language?

septag:
The problem with windows SDK is that they have these weird namings and versioning added to their path.
for example on my machine, there are 3 SDKs (8.0, 8.1, 10), and SDK for 10, it has 4 different versions, something like this: 
_C:\Program Files (x86)\Windows Kits\10\Lib_ : 
```
10.0.14393.0
10.0.10586.0
10.0.10240.0
10.0.10150.0
```
and an _um_ added to their include/ paths

So I had to get their directory names and sort them so I could get the latest version.
But of course you are right about the approach to the problem, I have found a cmake-module that queries registery for install path of the SDK and gives me these directories much nicer, I'll commit tommorow.

septag:
Fixed the cmake module, check it out

cmake/FindDSound.cmake
```diff
@@ -8,68 +8,28 @@
...
-# DSOUND_INCLUDE_DIR
-file(GLOB DXSDK_INCLUDE_DIRS LIST_DIRECTORIES TRUE "C:/Program Files (x86)/Windows Kits/10/Include/*")
-if (DXSDK_INCLUDE_DIRS)
-  list(SORT DXSDK_INCLUDE_DIRS)
-  list(REVERSE DXSDK_INCLUDE_DIRS)
-endif() 
+if (WIN32)
+  include(FindWindowsSDK)
+  get_windowssdk_library_dirs(${WINDOWSSDK_PREFERRED_DIR} WINSDK_LIB_DIRS)
+  get_windowssdk_include_dirs(${WINDOWSSDK_PREFERRED_DIR} WINSDK_INCLUDE_DIRS)
```
kcat: This looks like it should check ```WINDOWSSDK_FOUND``` before making those calls, otherwise those functions aren't defined.

septag: fixed

PR #80, opened 2016-11-04
SOFA importer for makehrtf

hoene:


CMakeLists.txt
```diff
@@ -1336,74 +1333,72 @@ IF(ALSOFT_UTILS)
...
+	    	utils/sofa/SOFASimpleFreeFieldHRIR.cpp
+	    	utils/sofa/SOFASimpleFreeFieldSOS.cpp
+	    	utils/sofa/SOFASimpleHeadphoneIR.cpp
+	    	utils/sofa/SOFAGeneralTF.cpp
+	    	utils/sofa/SOFAGeneralFIR.cpp
+	    	utils/sofa/SOFASource.cpp
+	    	utils/sofa/SOFAString.cpp
+	    	utils/sofa/SOFAUnits.cpp
+	    	utils/sofa/SOFAtoOpenAL.cpp
+	    	PROPERTIES COMPILE_FLAGS " -DLINUX=1 -D__unix__=1 -DHOST_ENDIAN_LITTLE=1 -Wno-ignored-qualifiers")
```
kcat: Unfortunately, things like this will prevent this from being accepted. It presumes a specific target system, but it should set the needed macros depending on the actual target as reported by cmake (e.g. it should be able to work with big-endian and/or non-Linux system). Consequently, hard-coding linking to ```libnetcdf_c++4.so``` won't work either, since different systems may dress their libs differently.

Also, it shouldn't be cmake's job to download and build netcdfc++, as that's a very system-dependent operation (not all systems will have wget, or can use configure scripts). It should just check for it in the normal way, using a Find* style module that allows specifying a custom location through a ```*DIR``` variable, leaving it to the user to ensure it's already built and set up properly.

choene:
A vanilla SOFA file reader
Supports reading many of the available SOFA files.

choene:
Not all SOFA files can be used for OpenAL-Soft

Some limitations are:
- The filter must be organized in a grip with constant elevation angles over horizontal azimuth angles.
- Only a limited number of different elevation angles is supported. Same is true for the maximal number of azimuth angle per elevation.
- Different radii are not support

kcat:
Support for SOFA files has been added with commit ecafa19bcfc2a4d27183b40066de163cb97dbdbf, using libmysofa. It's not as simple as passing a sofa file to makehrtf -- it still needs a def file ([example](https://github.com/kcat/openal-soft/blob/master/utils/SCUT_KEMAR.def) that specifies certain properties and a uniform distribution of depth, elevation, and azimuth positions to use -- but it is much simpler than having to provide and list hundreds or thousands of individual files.

There's still room for improvement (getting properties it can from the sofa file instead of relying on the def file for them, automatically deriving a uniform distribution to use like `sofa-info` does), but basic support is there now.

hoene:
nice!

PR #85, opened 2016-11-24
Add logcat logging

CAMOBAP795:


kcat:
I don't suppose ```__android_log_print``` also forwards the message to ```stderr```? Or is the output only visible afterward using logcat?

CAMOBAP795:
@kcat 

> I don't suppose __android_log_print also forwards the message to stderr

Yes, because [`"By default, the Android system sends stdout and stderr (System.out and System.err) output to /dev/null"`](https://developer.android.com/studio/command-line/logcat.html#viewingStd)`

Buy some devices may send `stderr` to `logcat` (for example my Samsung S3 did this)

https://developer.android.com/studio/command-line/logcat.html#viewingStd - here is more details how to forward `std***` to `logcat`

kcat:
> But some devices may send ```stderr``` to ```logcat``` (for example my Samsung S3 did this)
>
> https://developer.android.com/studio/command-line/logcat.html#viewingStd - here is more details how to forward ```std***``` to ```logcat```

That sounds like a more appropriate solution. Have Android forward ```stderr``` to the system log if you want the messages there, instead of introducing special logging behavior for one system that would prevent users from redirecting OpenAL Soft's logging output as documented. Is there much benefit to having OpenAL Soft write its log to the system log explicitly, instead of enabling/redirecting stderr to somewhere visible?

I could perhaps add an option that writes to both ```stderr``` and the system log on Android, though.

CAMOBAP795:
> That sounds like a more appropriate solution.
Your are right, but the problem that in reality this feature work unstable, sometimes require root access

Several posts about it:
https://groups.google.com/forum/#!msg/android-ndk/teb3fbYc2XQ/3nzf9k2yNw0J
http://stackoverflow.com/questions/10531050/redirect-stdout-to-logcat-in-android-ndk
http://stackoverflow.com/questions/5499202/why-is-redirecting-stdout-stderr-on-android-not-working

Even for case if redirecting work `__android_log_print` have some advantages:

1. Log level explicitly specified and it can be used during filtering (for `str**` you have only two log levels `E` and `D`
2. Log tag explicitly specified and it can be used during filtering, for `std**` you will see tags `System.err` and `System.out`,

kcat:
Commit 3ba03c5a29bcebb1866f36555957c3c293a99962 (cleaned up a bit with 9c019126d4d7450e0dc12f48824fb8c1214ba545) adds logcat logging on top of the normal stderr output. It will always make the __android_log_print calls regardless of the logging level, under the assumption that Android will suppress DEBUG and WARN log statements under normal use (though it shouldn't be a big deal even if not; such calls aren't too pervasive, particularly the latter).

If there's still any issues with it, feel free to either make another pull request, or open an issue for it.

PR #96, opened 2017-03-05
Remove duplicate CMake code when adding backends.

adrianbroher:
The PR moves some duplicated CMake code like `option` definitions, `set`-ting variables or checking for availability of a specific backend to a common CMake macro.  As far as I could check this seems to work properly for the backends *ALSA, OSS, PORTAUDIO, PULSEAUDIO, JACK* but I haven't no ways to test the other backends, maybe the Travis/AppVeyor can give me some answers here.


adrianbroher:
Looks like WinMM, DSound and MMDevAPI are detected propertly, but the TravisCI build almost enables no backends because the dependencies are not installed.  Is this intended?

CMakeLists.txt
```diff
@@ -867,209 +906,60 @@ SET(ALC_OBJS  ${ALC_OBJS}
...
-        ADD_BACKEND_LIBS(${ALSA_LIBRARIES})
-        IF(CMAKE_VERSION VERSION_LESS "2.8.8")
-            INCLUDE_DIRECTORIES(${ALSA_INCLUDE_DIRS})
-        ENDIF()
-    ENDIF()
-ENDIF()
-IF(ALSOFT_REQUIRE_ALSA AND NOT HAVE_ALSA)
-    MESSAGE(FATAL_ERROR "Failed to enabled required ALSA backend")
+ADD_BACKEND(ALSA ALSA Alc/backends/alsa.c)
+IF(ALSOFT_BACKEND_ALSA)
```
kcat: I'm not sure these should check the ```ALSOFT_BACKEND_*``` options, since they may be cached values and will stay even if the backend is no longer available. I think it makes more sense to check the ```HAVE_*``` variables.

cmake/FindWinMM.cmake
```diff
@@ -0,0 +1,39 @@
+# - Find Windows Multi Media Extensions (WinMM) includes and libraries
+#
+#   WINMM_FOUND        - True if WINMM_INCLUDE_DIR & WINMM_LIBRARY are found
+#   WINMM_LIBRARIES    - Set when WINMM_LIBRARY is found
+#   WINMM_INCLUDE_DIRS - Set when WINMM_INCLUDE_DIR is found
+#
+#   WINMM_INCLUDE_DIR - where to find dsound.h, etc.
+#   WINMM_LIBRARY     - the dsound library
```
kcat: Seems to be a bit of a copy-paste error here.

kcat:
> Looks like WinMM, DSound and MMDevAPI are detected propertly, but the TravisCI build almost enables no backends because the dependencies are not installed. Is this intended?

I don't really know. I'm not familiar with using TravisCI, so I'm not sure if I can or how to enable any dependencies.

adrianbroher:
> I'm not sure these should check the ALSOFT_BACKEND_* options, since they may be cached values and will stay even if the backend is no longer available. I think it makes more sense to check the HAVE_* variables.

Yeah, checking against a user option isn't the best idea. I replaced the `ALSOFT_BACKEND_*` variables with `HAVE_*` counterparts.

> Seems to be a bit of a copy-paste error here.

Yes, it was a copy-paste error.

kcat:
Seems the Android build fails, unable to find the OpenSLES lib: https://travis-ci.org/kcat/openal-soft/jobs/208520864. Can't exactly see why though, since it seems to be looking for the same lib name. Perhaps a default search path isn't being set for ```find_library```?

PR #97, opened 2017-03-05
Enable and enforce dependencies on CI services.

adrianbroher:
This PR is a work in progress

This PR (should) contain commits to allow that more parts of the project like backends or example programs are build and enforced to build to check for regressions in the build system.  Due to the nature of the CI services in question I can only test changes as external contributor by pushing to to this PR.  I will notify you when I consider the changes good enough for a review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,16 +27,18 @@ install:
     fi
   - >
     if [[ "${TRAVIS_OS_NAME}" == "linux" && "${BUILD_ANDROID}" == "true" ]]; then
-      curl -o ~/android-ndk.zip https://dl.google.com/android/repository/android-ndk-r14-linux-x86_64.zip
-      unzip -q ~/android-ndk.zip -d ~ \
-        'android-ndk-r14/build/cmake/*' \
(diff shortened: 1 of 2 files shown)


adrianbroher:
After some tuning the CI now builds all backends.

kcat:
So this is now ready for merge, then?

adrianbroher:
> So this is now ready for merge, then?

Yes, unless you have questions or remarks to the PR itself.

kcat:
Only thing I notice is that the Android build doesn't build with NEON support. The ```-mfpu=neon``` check passes, but it seems to not find ```arm_neon.h```. It's not too big of a deal, but do you know why that might be?

adrianbroher:
> Only thing I notice is that the Android build doesn't build with NEON support.

I'm not well versed with embedded hardware, but the NDK toolchain files seems to select the `armeabi-v7a` without NEON support if you don't add any specific CMake parameters.  Maybe the cross compiler always supports `-mfpu=neon` and therefor is not a reliable check.

https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake

PR #131, opened 2017-06-28
Suggestion: Clarify float format symbols

ColdPie1:
I won't be offended if you close this with "get a clue" ;)

I find it confusing that the integer formats consistently are untyped (AL_FORMAT_QUAD8, AL_MONO8_SOFT, AL_FORMAT_BFORMAT2D_8) and float formats are consistently typed (AL_MONO32F_SOFT, AL_FORMAT_BFORMAT2D_FLOAT32) except in the case of the multichannel extension, where the float format is untyped (AL_FORMAT_QUAD32). This led me to reject what was probably a perfectly fine patch to Wine, see https://bugs.winehq.org/show_bug.cgi?id=42414

There's some evidence that I'm not the only clueless one:

http://openal.org/pipermail/openal/2014-December/000286.html

https://developer.xamarin.com/api/type/OpenTK.Audio.OpenAL.ALFormat/

This PR has two commits, either of which alone would solve the problem, so both may be overkill. The first simply adds a comment to the header explicitly explaining each format. The second adds "32F" style aliases to the existing "32" symbols. You have a better picture of the pros/cons of adding new symbols than I do.

### DIFF
--- a/include/AL/alext.h
+++ b/include/AL/alext.h
@@ -97,6 +97,17 @@ extern "C" {
 
 #ifndef AL_EXT_MCFORMATS
 #define AL_EXT_MCFORMATS 1
+/*
+ * QUAD -> 4.0 quad (FL, FR, BL, BR)
+ * REAR -> rear stereo (BL, BR)
+ * 51CHN -> 5.1 (FL, FR, FC, LFE, SL, SR)
+ * 61CHN -> 6.1 (FL, FR, FC, LFE, BC, SL, SR)
+ * 71CHN -> 7.1 (FL, FR, FC, LFE, BL, BR, SL, SR)
+ *
+ * 8 -> unsigned 8-bit integer
+ * 16 -> signed 16-bit integer
+ * 32 -> IEEE 32-bit float
+ */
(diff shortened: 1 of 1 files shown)


kcat:
I don't think more enums should be added to public headers without an associated extension, since the app should be able to query them using ```alGetEnumValue```.

Properly documenting things in ```alext.h``` isn't a bad idea, though.

ColdPie1:
Would you like me to update the PR? Are you happy with the comment I added?

kcat:
I think a little more detail would be better. Describing the channel order for the formats as being the same as WaveFormatEx, for example, along with the sample types.

ColdPie1:
I've updated the branch to improve the comment. I'm not sure about the REAR format. It doesn't seem to be handled anywhere in the code.

kcat:
The multichannel formats now have short descriptions in `alext.h`, with commit f48a1e3a314dcbaf8b96d44be17c8fb708d35ae8. Does that work?

ColdPie1:
Yes, that clears up everything I was confused about. Thanks!

PR #139, opened 2017-08-08
Restore the JACK error message handler in a JACK1 compatible manner

### LINKED ISSUE #138: Crash when Jack backend is chosen
```jack_activate(self->Client)``` called in ```ALCjackPlayback_start``` crashes.

Other apps using Jack, like ```ardour5```, work fine.

Observed on the FreeBSD 11.1, jack-0.125.0, openal-soft-1.18.0.

### PR DESCRIPTION
**Pull request by kdhp**
JACK1 does not allow `NULL` to be passed to `jack_set_*_function`, while JACK2
restores the default callback.  This change explicitly resets the error function to `default_jack_error_callback`, as provided by the library, and is compatible with
both versions.

Fixes #138

### DIFF
--- a/Alc/backends/jack.c
+++ b/Alc/backends/jack.c
@@ -39,6 +39,16 @@ static const ALCchar jackDevice[] = "JACK Default";
 
 
 #ifdef HAVE_DYNLOAD
+
+/*
+ * `default_jack_error_callback' is exported by JACK1, while in JACK2
+ * the callback is reset to the default by passing a NULL to
+ * `jack_set_error_function'.  As such, after loading this symbol, passing
+ * the resulting pointer to `jack_set_error_function' will reset the error
+ * callback regardless of the version of JACK used.
+ */
+void (*pdefault_jack_error_callback)(const char *);
+
 #define JACK_FUNCS(MAGIC)          \
     MAGIC(jack_client_open);       \
     MAGIC(jack_client_close);      \
@@ -54,6 +64,7 @@ static const ALCchar jackDevice[] = "JACK Default";
     MAGIC(jack_get_ports);         \
     MAGIC(jack_free);              \
     MAGIC(jack_get_sample_rate);   \
+    MAGIC(jack_error_callback);    \
     MAGIC(jack_set_error_function); \
     MAGIC(jack_set_process_callback); \
     MAGIC(jack_set_buffer_size_callback); \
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by kcat** [comment]
Won't this fail if ```default_jack_error_callback``` isn't exported? It doesn't look like JACK2 exports it, so the lib will fail to load dynamically at runtime (```GetSymbol```/```dlsym``` will return ```NULL```, causing the load to abort since it didn't get an expected symbol).

**Comment by kdhp** [comment]
@kcat I may have misread, however `jack_error_callback` is explicitly exported by both versions, so that can be safely used instead.

**Comment by kcat** [comment]
It doesn't seem that JACK2 exports ```jack_error_callback```:
```
$  nm -D /usr/lib/x86_64-linux-gnu/libjack.so.0.1.0 | grep callback
000000000000f2e0 T jack_client_has_session_callback
000000000000d2f0 T jack_set_buffer_size_callback
000000000000d390 T jack_set_client_registration_callback
000000000000d1e0 T jack_set_freewheel_callback
000000000000d4d0 T jack_set_graph_order_callback
000000000000d590 T jack_set_latency_callback
000000000000d430 T jack_set_port_connect_callback
000000000000d3e0 T jack_set_port_registration_callback
000000000000d480 T jack_set_port_rename_callback
000000000000d060 T jack_set_process_callback
000000000000f3d0 T jack_set_property_change_callback
000000000000d340 T jack_set_sample_rate_callback
000000000000eec0 T jack_set_session_callback
000000000000e420 T jack_set_sync_callback
000000000000d5e0 T jack_set_thread_init_callback
000000000000e4c0 T jack_set_timebase_callback
000000000000d540 T jack_set_xrun_callback
```
And:
```
AL lib: (WW) GetSymbol: Failed to load jack_error_callback: /usr/lib/x86_64-linux-gnu/libjack.so.0: undefined symbol: jack_error_callback
AL lib: (WW) jack_load: Missing expected functions:
jack_error_callback
AL lib: (WW) alc_initconfig: Failed to initialize backend "jack"
```

kdhp:
I seem to have stumbled back onto the JACK issue mentioned in #138.

`default_jack_error_callback` is exported by JACK1, while in JACK2
the callback is reset to the default by passing a `NULL` to
`jack_set_error_function`.  As such, after loading this symbol, passing
the resulting pointer to `jack_set_error_function` will reset the error
callback regardless of the version of JACK used.

kdhp:
The 'Travis CI' build status seems to be stuck, on the 'Travis' site it is listed as successfully completed but it has not been updated here.

kcat:
This will still fail to load with JACK2 because ```jack_error_callback``` is missing. Also the check for ```default_jack_error_callback``` is done at build-time, so it will still fail if JACK2 is present at build-time but is later replaced with JACK1 when run (and vice-versa).

PR #143, opened 2017-08-31
Don't confuse MSVC with Clang on Windows

codicodi:
When building with Clang on Windows, `MSVC` evaluates to true (as Clang tries to be MSVC-compatible). To reliably check which compiler is used, one needs to test `CMAKE_C_COMPILER_ID`.
This patch fixes bunch of errors like this one:

```
G:\dev\tmp\openal-soft\Alc\mixer_sse41.c(124,15):  error: always_inline function '_mm_hadd_ps' requires target feature 'sse3',
 but would be inlined into function 'Resample_fir4_32_SSE41' that is compiled without support for 'sse3'
        out = _mm_hadd_ps(k0, k2);
              ^
```
Note that there is no need to change anything in other cases, such as here: https://github.com/kcat/openal-soft/blob/0b0ae75ccf684f850f9112dbf17072b0c22da5f6/CMakeLists.txt#L292-L315. Clang will happily accept `/MT` and similar flags


kcat:
> When building with Clang on Windows, MSVC evaluates to true

Sounds like a number of nasty problems waiting to happen, especially if it still needs those GCC switches to handle SSE intrinsics. Some beneficial GCC switches are excluded with MSVC, and there's workarounds/fallbacks in the source for MSVC that could be detrimental to take if not otherwise needed.

In this particular case the MSVC check wasn't necessary, so I removed it with commit 653edd4b02ce9ea7eaf49a51be0b27c18cd3db63. But if Clang is reporting itself as MSVC despite acting like not-MSVC, the generated lib might not be as optimized as it could be.

PR #157, opened 2017-11-26
Do not probe a device which backend not fully implemented.

devnexen:
Spotted with backend without capture device handling.


kcat:
This can actually leak the list lock (it takes the lock, then returns without releasing the lock if there's no backend). I've implemented an alternate fix for this in commit d96be76875fd93da8ecb0d7a7096614042e3e02a.

kcat:
Or I should say, commit 5993ef0baa9096ca62d8bca0f17b95c5468ffde3.

PR #160, opened 2017-12-24
Add Libs.private to openal.pc

shinchiro:
Necessary for proper static linking


CMakeLists.txt
```diff
@@ -1229,6 +1229,11 @@ SET(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
 SET(PACKAGE_VERSION "${LIB_VERSION}")
 IF(LIBTYPE STREQUAL "STATIC")
     SET(PKG_CONFIG_CFLAGS -DAL_LIBTYPE_STATIC ${PKG_CONFIG_CFLAGS})
+    FOREACH(LIB ${EXTRA_LIBS})
+        IF(NOT LIB MATCHES "^-.*")
+            SET(PKG_CONFIG_LIBS_PRIVATE "${PKG_CONFIG_LIBS_PRIVATE} -l${LIB}")
```
kcat: As far as I can see, this will add ```-lwhatever``` for every entry in ```EXTRA_LIBS``` that doesn't start with ```-```. Is that correct? If so, I'm not sure it'll work correctly for entries that have a full path, such as the frameworks on OSX, or the result of the ```Find*``` scripts.

shinchiro: > As far as I can see, this will add -lwhatever for every entry in EXTRA_LIBS that doesn't start with -. Is that correct?
Yes. Probably it's better to left out the if check?

kcat: > Also does libraries return by Find* scripts include -l flags?

No, they give the full path and filename for the library (the base ```.so``` symlink, or the ```.a```, ```.dylib```, ```.framework```, ```.dll.a```, or ```.lib```, depending on the system and compiler). I honestly don't know what the correct way to handle it would be.

kcat:
Should be handled now with commit a2c24ff946623b74df6ca7884f549a76911ebe8b. Just be aware that when CMake searches for libraries it provides them as a full path+filename, which get passed to `Libs.private` as-is rather than as -L and -l switches. So certain optional dependencies may rely on specific paths if you intend to transfer the `pc` file to other machines.

If there's any issues, feel free to let me know.

shinchiro:
Works fine in my case

PR #167, opened 2018-01-28
Add option to disable "build_version" cmake target

aaronmjacobs:
When using openal-soft as a git submodule, the "build_version" target currently gets generated. This causes any cmake targets that link against openal-soft to request a build every time they are launched in Visual Studio. I've added a new cmake option to allow disabling the creation of the "build_version" target to get around this issue.


kcat:
That's odd. Why does it request a build when nothing's changed? There are other places `ADD_CUSTOM_TARGET` is used which can't be disabled. Don't they cause it as well?

aaronmjacobs:
I currently using openal-soft version 1.18.2, which only has one custom target - I would assume the same issue exists with any other custom targets that are in the master branch. According to the [cmake documentation](https://cmake.org/cmake/help/latest/command/add_custom_target.html), add_custom_target:

> Adds a target with the given name that executes the given commands. The target has no output file and is always considered out of date even if the commands try to create a file with the name of the target.

PR #175, opened 2018-03-07
Add SDL2 backend for playback, fix #173

### LINKED ISSUE #173: SDL2 as a backend
I was wondering if it would be easy to use SDL2 as a *backend*? I'm currently using OpenAL for a game and SDL2 is available for some game consoles where none of the other backends have been ported yet.

jhasse:
I wasn't sure about coding style, so I mostly took `backends/null.c` and `backends/wave.c` as a reference. Looking forward to your feedback :)

Capture support can be added later, too. I don't need it though.


kcat:
For an SDL2 backend, it would probably make more sense to use its callback rather than using a timing loop. Especially since the system clock and audio clock aren't necessarily synchronized. The PortAudio backend uses a method like that.

It may also be a good idea to not force the SDL2 device to the specific format. SDL2 will return the format it actually sets, and the ALCdevice's format can be updated to match. Also the reset method should fix up the ALCdevice's format since it may be requesting something different than what it opened with, and SDL2 apparently can't change the actual device format afterward. And don't forget to make the stop method pause the SDL2 device.

jhasse:
> For an SDL2 backend, it would probably make more sense to use its callback rather than using a timing loop. Especially since the system clock and audio clock aren't necessarily synchronized. The PortAudio backend uses a method like that.

Makes sense! I've switched to the callback method and the code is a lot shorter now, too :)

> It may also be a good idea to not force the SDL2 device to the specific format. SDL2 will return the format it actually sets, and the ALCdevice's format can be updated to match.

I've passed `0` so that SDL2 isn't allowed to do that. I've added a commit which would allow frequency changes, but for format changes, I don't know enough about OpenAL, SDL2, audio in general to be confident enough to implement it.

> Also the reset method should fix up the ALCdevice's format since it may be requesting something different than what it opened with, and SDL2 apparently can't change the actual device format afterward.

Not sure I understand. Do you mean that I should save `SDL_AudioSpec` and set the device settings to its values in `..._reset`?

> And don't forget to make the stop method pause the SDL2 device.

Done, thanks :)

kcat:
> for format changes, I don't know enough about OpenAL, SDL2, audio in general to be confident enough to implement it.

For the channel configuration, if SDL2 gives back `1` it's `DevFmtMono`, and `2` it's `DevFmtStereo`. The structure definition says those are the only two valid values for `channels`, but I don't know if it's an undocumented "feature" to be something else, like 4 for quad or 6 for 5.1. If it can be other things, that would indeed complicate things.

For the sample format, it can be mapped as: `AUDIO_U8` -> `DevFmtUByte`, `AUDIO_S8` -> `DevFmtByte`, `AUDIO_U16SYS` -> `DevFmtUShort`, `AUDIO_S16SYS` -> `DevFmtShort`, `AUDIO_S32SYS` -> `DevFmtInt`, `AUDIO_F32SYS` -> `DevFmtFloat`. Endian mismatches (e.g. AUDIO_S16LSB on a big endian system) would need appropriate byte swapping.

> Not sure I understand. Do you mean that I should save SDL_AudioSpec and set the device settings to its values in ..._reset?

Yes.

jhasse:
Thanks, tried to implement it the best I could, please have another look.

> Endian mismatches (e.g. AUDIO_S16LSB on a big endian system) would need appropriate byte swapping.

I don't know how to do that. Also I'm not really interested in implementing it, all the other backends are there after all ;)

jhasse:
Thanks for the merge and also big thanks for the review :)

PR #177, opened 2018-03-18
EFX:Pitch shifter

Raulshc:
### PROPOSE
This pull request implements the pitch shifter effect of ALC_EXT_EFX extension. Standard phase vocoder is the used algorithm.

### TODO

- [x] Pitch shifter implementation into ALC_EXT_EFX 
- [x] Add support to alsoft-config for handle it.

Review and merge it,  please
Thanks


kcat:
How much does the window size depend on the output quality? Since `BUFFERSIZE` is set for the mixing size (currently 2048, but could be set lower in the future), it might not be the best to use for the FFT. Also, it's probably not a good idea to place `bufferOut` on the stack if it's going to be `BUFFERSIZE` samples.

Raulshc:
> How much does the window size depend on the output quality? Since `BUFFERSIZE` is set for the mixing size (currently 2048, but could be set lower in the future), it might not be the best to use for the FFT.

Of course, the STFT window size and oversampling (overlap) coefficient depends on output quality. Typical values for a normal quality are: 1024 for window size and 4 (75%) for oversampling(overlap) factor. A good quality has the values: 2048, 4.

I think it's right that the value of the STFT window isn't linked to the mix size. I took it because I thought it was a stable value and that it would not be modified in the future, and it matches the value needed for good quality. I will make a commit with this small change.

> Also, it's probably not a good idea to place `bufferOut` on the stack if it's going to be `BUFFERSIZE` samples.

The current structure of the algorithm is designed using FIFO buffers, this allows you to process an arbitrary number of samples, regardless of the window size, giving output the same number of samples. As the number of samples is variable, we only have to know the maximum possible dimension for BufferOut.

I understand that the maximum possible dimension is `BUFFERSIZE` and that's why I initialize that vector with `BUFFERSIZE` elements.

This may not be the best, but you can change it later with some method that you think is more convenient. I have done tests with several combinations of `period_size` and `periods` using alsoft.ini and the sound behaves in a stable way.

Thanks,
Regards

PR #178, opened 2018-03-19
Examples: Fix MSVC link error

Raulshc:
### PROPOSE
MSVC prior to 2013  does not include `strtof ` in `<stdlib.h>`.  This commit fix this issue for ms compilers prior to VS2013.

See strtof in MSDN: [https://msdn.microsoft.com/en-us/library/dn320175(v=vs.140).aspx](https://msdn.microsoft.com/en-us/library/dn320175(v=vs.140).aspx)

Review it and merge, please.
Thanks

### DIFF
--- a/examples/alrecord.c
+++ b/examples/alrecord.c
@@ -184,7 +184,7 @@ int main(int argc, char **argv)
                 return 1;
             }
 
-            recorder.mRecTime = strtof(argv[1], &end);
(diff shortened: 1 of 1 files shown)


kcat:
This was fixed a different way in commit 55a73b8df94ce3abc4d6c6a654e2516acf8ef631.

PR #193, opened 2018-05-20
EFX: Frequency shifter

Raulshc:
### PROPOSE
This pull request implements the frequency shifter effect of ALC_EXT_EFX extension. The algorithm uses discrete Hilbert transform. Only nono source proccessed by now (left direction)
### TODO
- [x] Frequency shifter implementation into ALC_EXT_EFX 
- [x] Add support to alsoft-config for handle it.

Review and merge it, please
Thanks


kcat:
Out of curiosity, what is this effect actually intended to do? The description in the EFX docs is rather sparse:

> The frequency shifter is a single-sideband modulator, which translates all the component frequencies of the input signal by an equal amount.

The Left Direction and Right Direction parameters are also extremely light on their descriptions. "These select which internal signals are added together to produce the output.combinations of values will produce slightly different tonal and spatial effects." doesn't say a whole lot. It doesn't seem like it means the left and right input/output, but I can't say for sure.

Raulshc:
The frequency shifter is an implementation of single-sideband modulation (SSB), it's described, for example in:

[https://en.wikipedia.org/wiki/Single-sideband_modulation](url)

It's similar to Amplitude modulation (ring modulation in the oal case), with the particularity that only one sideband of the output signal spectrum is preserved. (BTW, ring modulator have a slight bug that we would to correct).

Using Creative Host OpenAL implementation (software) as blackbox test reference, I saw that `AL_FREQUENCY_SHIFTER_LEFT_DIRECTION` implements the processed signal on the left channel (mono), `AL_FREQUENCY_SHIFTER_RIGHT_DIRECTION` is not in use, `AL_FREQUENCY_SHIFTER_DIRECTION_DOWN` corresponding to the USB (upper sideband) election and `AL_FREQUENCY_SHIFTER_DIRECTION_UP` corresponding to LSB (upper sideband). Creative Host OpenAL implementation doubles the carrier frequency too (by mistake, I think), unlike this implementation.

I have a creative X-Fi Fatal1ty to be used as (hardware) blackbox test reference, but I haven't installed in a PC yet, so I don't know its behavior in the tests of the different effects.

kcat:
I see. I wouldn't be surprised if Creative's software driver uses mono inputs for effects, so only one channel ("left") is affected. For OpenAL Soft it wouldn't be hard to support left and right processing, but that can be done later.

PR #201, opened 2018-06-08
Fix missing second level sysctl(3) node

kdhp:
Add the missing second level node `KERN_PROC` that is required for the `KERN_PROC_ARGS` sysctl(3) calls to succeed.

### DIFF
--- a/Alc/helpers.c
+++ b/Alc/helpers.c
@@ -676,13 +676,13 @@ void GetProcBinary(al_string *path, al_string *fname)
     size_t pathlen;
 
 #ifdef __FreeBSD__
-    int mib[4] = { CTL_KERN, KERN_PROC_ARGS, getpid() };
-    if(sysctl(mib, 3, NULL, &pathlen, NULL, 0) == -1)
-        WARN("Failed to sysctl kern.procargs.%d: %s\n", mib[2], strerror(errno));
+    int mib[4] = { CTL_KERN, KERN_PROC, KERN_PROC_ARGS, getpid() };
+    if(sysctl(mib, 4, NULL, &pathlen, NULL, 0) == -1)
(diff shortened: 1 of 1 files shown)


kdhp:
Relevant table from the [FreeBSD man-page](https://www.freebsd.org/cgi/man.cgi?query=sysctl&sektion=3):

```
                   Third Level Name          Fourth Level
                   KERN_PROC_ARGS            A process ID
                   KERN_PROC_PATHNAME        A process ID
```

kcat:
I'm curious about the difference between `KERN_PROC_ARGS` and `KERN_PROC_PATHNAME`. By the sounds of it, `KERN_PROC_ARGS` returns the (null separated) list of arguments. Is it safe to assume argv[0] would contain the full absolute path and filename for the process? Or might it be a relative path, or have no path at all?

For `KERN_PROC_PATHNAME`, what's the process' text file? Would that be the binary executable, or something else? And would it contain the absolute path?

kdhp:
`KERN_PROC_ARGS` essentially returns a copy of the initial `argv` array, and `argv[0]` can not be relied on because [it can be set arbitrarily on execution](https://www.freebsd.org/cgi/man.cgi?query=execve):

> At least one argument must be present in the array; by custom, the first element
> should be the name of the executed program (for example, the last component of path).

The "process' text file" [does refer to its executable](https://svnweb.freebsd.org/base/head/sys/sys/proc.h?revision=334595&view=markup#l626), and `KERN_PROC_PATHNAME` [does return the full path](https://svnweb.freebsd.org/base/head/sys/kern/kern_proc.c?revision=332122&view=markup#l2103), if possible.

kcat:
I fixed it up to use `KERN_PROC_PATHNAME` instead, to more reliably get the full path and executable. If there's still an issue with it, feel free to make a new pull request.

PR #204, opened 2018-06-19
Correctly check byte order for newer Android compiler, fix #203

### LINKED ISSUE #203: Android/OpenSL: unsupported byte order 1
I've cross-compiled OpenAL Soft from Linux for Android like this:

```
cd build
cmake -DCMAKE_SYSTEM_NAME=Android -DANDROID_NDK=$HOME/Android/Sdk/ndk-bundle/ \
    -DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a -DALSOFT_BACKEND_OSS=OFF \
    -DCMAKE_ANDROID_NDK_TOOLCHAIN_VERSION=clang -DALSOFT_BACKEND_WAVE=OFF ..
make
```

Trying to play a sound on a LG G5 (Android 7.0) `alcCreateContext` returns `NULL`. Looking at the logs I've found:

```
06-19 09:58:38.128 22245 22262 D openal  : AL lib: UpdateDeviceParams: Pre-reset: Stereo, Float, 44100hz, 1024 update size x3
06-19 09:58:38.129 22245 22262 E libOpenSLES: Cannot create audio player: unsupported byte order 1
06-19 09:58:38.129 22245 22262 W libOpenSLES: Leaving Engine::CreateAudioPlayer (SL_RESULT_CONTENT_UNSUPPORTED)
06-19 09:58:38.129 22245 22262 E openal  : AL lib: ALCopenslPlayback_reset: engine->CreateAudioPlayer: Content unsupported
```

Byte order 1 seems to be `SL_BYTEORDER_BIGENDIAN`. If I manually add

```
format_pcm.endianness = SL_BYTEORDER_LITTLEENDIAN;
```

before the call to `CreateAudioPlayer` in `opensl.c`, `alcCreateContext` succeeds and I can play sounds. If I understand it correctly `armeabi-v7a` is little endian, why does `IS_LITTLE_ENDIAN` seem to be false?

jhasse:
Shortly after I've opened the issue, I found the cause for the wrong return value of `IS_LITTLE_ENDIAN`: `__BYTE_ORDER__` is set to `__ORDER_LITTLE_ENDIAN__` instead of just `__LITTLE_ENDIAN__` (Android NDK r17 btw).

### DIFF
--- a/OpenAL32/Include/alMain.h
+++ b/OpenAL32/Include/alMain.h
@@ -184,8 +184,8 @@ inline int fallback_ctz64(ALuint64 value)
 #define CTZ64 fallback_ctz64
 #endif
 
-#if defined(__BYTE_ORDER__) && defined(__LITTLE_ENDIAN__)
-#define IS_LITTLE_ENDIAN (__BYTE_ORDER__ == __LITTLE_ENDIAN__)
+#if defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__)
+#define IS_LITTLE_ENDIAN (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
 #else
 static const union {
     ALuint u;


kcat:
Actually it seems there was a mistake on my part and it shouldn't be comparing `__BYTE_ORDER__` to `__LITTLE_ENDIAN__`. According to https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html it should be using `__ORDER_LITTLE_ENDIAN__` instead.

jhasse:
If I understand it correctly `__LITTLE_ENDIAN__` is a macro on it's own which is defined, if `__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__`.

I will update the PR.

kcat:
I don't think we need to worry about `__LITTLE_ENDIAN__` with `__BYTE_ORDER__` being used properly. Less risk of compatibility issues if different compilers or systems handle the `__LITTLE_ENDIAN__` macro differently.

jhasse:
Okay, removed that :)

PR #212, opened 2018-07-21
EFX: Modulator fixes #2

### LINKED ISSUE #2: Alc/mixer_neon.c typo
void MixDirect_Neon() should be void Mix_Neon()

Raulshc:
This pull request fixes the max limit of the normalized frequency in the biquad filter. The normalized frequency must be below to 0.5; with `f0norm=0.49f`,the filter behaves stable. The PR fixes the modulator output with carrier signal frequency value = 0Hz too.

Review and merge it, please
Thanks


kcat:
My main concern is the discontinuity created by forcing the modulation value to 1 from wherever it was on the previous run. The ideal response would probably be for the modulator to continue as it was until it crosses 1, but that would certainly be more difficult to do and may still have issues (e.g. if the previous run just passed 1 and had a really low non-0 carrier frequency; it could take several seconds to reach 1 again). Alternatively, it could track the last generated sample, and increase the value back up to 1 over a number of samples (fast enough to not feel a delay, but slow enough to prevent unnecessary noise). Perhaps it's not a big deal to be like you have it, though. 

Another less pressing issue is the `if(!step)` check in the modulator loop. The value of `step` won't change in between calls to the `update` method, so you could just check it once before the loop and fill the output with `1.0f` in that case, else do the loop normally. Alternatively, there could be a separate `ModulateOne` function that just generates `1.0f`, and is selected in the `update` method instead of one of the other `Modulate*` functions when `step` ends up as 0.

Raulshc:
Making tests with the modulator I saw this concretely. For example, if you play a sound file using ring modulator effect with a sampling frequency of 44100Hz, the frequency parameter assigned to `AL_RING_MODULATOR_FREQUENCY` equal to 0 Hz with a waveform `AL_RING_MODULATOR_SINUSOID` from the beginning of the file, oalsoft assigns a step = 1 in the signal generator, a 24-bit `WAVEFORM_FRACBITS` resolution and therefore a `WAVEFORM_FRACONE` of 2^24. A step 1 with that resolution and sampling frequency corresponds to a period T=380.43s, and therefore a frequency of 0.026 Hz. In this way, the sound begins very low (it is multiplied by a sinus signal `sinf((ALfloat) index * (F_TAU / WAVEFORM_FRACONE))`), it have a maximum value in the t=95s and it would decay to zero in t=190s (3 min, 10 sec) and the cycle would be completed in the negative part of the signal.

> My main concern is the discontinuity created by forcing the modulation value to 1 from wherever it was on the previous run. The ideal response would probably be for the modulator to continue as it was until it crosses 1, but that would certainly be more difficult to do and may still have issues (e.g. if the previous run just passed 1 and had a really low non-0 carrier frequency; it could take several seconds to reach 1 again). Alternatively, it could track the last generated sample, and increase the value back up to 1 over a number of samples (fast enough to not feel a delay, but slow enough to prevent unnecessary noise). Perhaps it's not a big deal to be like you have it, though.

Don't worry with that, it's a simple multiplication by the neutral element of the product. It would be like turning off the modulator, it does not generate noise and gliches, you can check it for yourself.

> Another less pressing issue is the `if(!step)` check in the modulator loop. The value of `step` won't change in between calls to the `update` method, so you could just check it once before the loop and fill the output with `1.0f` in that case, else do the loop normally. Alternatively, there could be a separate `ModulateOne` function that just generates `1.0f`, and is selected in the `update` method instead of one of the other `Modulate*` functions when `step` ends up as 0.

As an optimization method, I like it more `ModulateOne` function that just generates `1.0f`.

Thank you very much
Regards

Raulshc:
All OK for me,

Thank you very much Chris,
Now, autowah effect ;)

PR #218, opened 2018-08-29
About compile on ios.

snowfoxer:
I try to compile the library in ios by XCode and test on our project.
Only include build_ios's openal-soft.xcodeproj in your own ios project. It will work.
The test is that I generate a serial of sin wav and put it to a loopback context, get it and play and finally put it into a file.
It works well.
I know that I am not familiar with the code, hope you can modify them and get the code run on ios.


kcat:
> Only include build_ios's openal-soft.xcodeproj in your own ios project.

It will only work for that particular configuration. CMake runs checks to find out what's actually available (or not available), and will generate the appropriate headers and project files from that. As it is, this is already out of date because the config.h is missing an definition that was just recently added for a function that is probably available, and version.h will have the wrong version (so any trace logs showing the version will be misleading).

What is the problem this is trying to solve?

kcat:
> I try to compile the library on ios and there's no cmake option about it.

It looks like [this toolchain](https://github.com/cristeab/ios-cmake) can be used to build a cmake project for an iOS target.

Wakever:
> > Only include build_ios's openal-soft.xcodeproj in your own ios project.
> 
> It will only work for that particular configuration. CMake runs checks to find out what's actually available (or not available), and will generate the appropriate headers and project files from that. As it is, this is already out of date because the config.h is missing an definition that was just recently added for a function that is probably available, and version.h will have the wrong version (so any trace logs showing the version will be misleading).
> 
> What is the problem this is trying to solve?

Wakever:
Where can I find the "build_ios's openal-soft.xcodeproj", please ?

PR #221, opened 2018-09-07
B-format support in HRTF example

karlun:
This pull request adds B-format support to the example file alhrtf.c by 1) adding three and four channel audio loading support, and 2) using source orientation to demonstrate rotating B-format playback.

It is unclear, however, how to control the channel order and normalization standard settings. I cannot find anything in the documentation, but my FOA FuMa file seems to play okay.

### DIFF
--- a/examples/alhrtf.c
+++ b/examples/alhrtf.c
@@ -301,6 +301,10 @@ int main(int argc, char **argv)
         /* This only rotates mono sounds. */
         alSource3f(source, AL_POSITION, (ALfloat)sin(angle), 0.0f, -(ALfloat)cos(angle));
 
+        /* This rotates b-format sources. */
+        ALfloat sourceOri[] = { (ALfloat)sin(-angle), 0.0f, -(ALfloat)cos(-angle), 0.0, 1.0, 0.0 };
+        alSourcefv(source, AL_ORIENTATION, sourceOri);
+
         if(has_angle_ext)
         {
             /* This rotates stereo sounds with the AL_EXT_STEREO_ANGLES
--- a/examples/alhrtf.c
+++ b/examples/alhrtf.c
@@ -89,6 +89,36 @@ static ALuint LoadSound(const char *filename)
             return 0;
         }
     }
+    else if(sample->actual.channels == 3)
+    {
+        if(sample->actual.format == AUDIO_U8)
+            format = AL_FORMAT_BFORMAT2D_8;
+        else if(sample->actual.format == AUDIO_S16SYS)
+            format = AL_FORMAT_BFORMAT2D_16;
+        else if(sample->actual.format == AUDIO_F32SYS)
+          format = AL_FORMAT_BFORMAT2D_FLOAT32;
+        else
+        {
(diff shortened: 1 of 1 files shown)


kcat:
It doesn't seem like SDL_sound can read B-Format [`amb` files](https://en.wikipedia.org/wiki/Ambisonic_data_exchange_formats#File_formats_and_metadata) for me. `Sound_NewSampleFromFile` returns `NULL` before the channel count can be checked. It seems it gets confused with the `WAVE_FORMAT_EXTENSIBLE` chunk, the `SUBTYPE_AMBISONIC_B_FORMAT_*` GUIDs, or because the channel mask is 0.

Are you using an actual amb file, or a plain 3- or 4-channel wav file that happens to contain B-Format audio? What version of SDL_sound are you using?

kcat:
Actually I take back that `Sound_NewSampleFromFile` fails. It actually succeeds. However, it reports only 2 channels for my amb file, which actually has 4 channels, and `Sound_DecodeAll` fails to read audio. There's also a bunch of console spam about failing to sync an mpeg stream.

karlun:
I used four channel wav file format. I have not tested amb files.

Also, I had to recompile SDL_sound against SDL2 since the SDL_sound version available on my Kubuntu release is compiled against SDL1 that does not support floating point waveform data. On the other hand, I had to convert my floating point wav file to an integer format anyway, since SDL_sound was unable to read floating point wav. The example supports float, however :-)

kcat:
I would be cautious with assuming random 3- and 4-channel wave files are B-Format, since there are "plain" 3-channel layouts (stereo+center, or stereo+LFE) and 4-channel layouts (quad). The reason for `amb` files (which are really wav files, with particular `WAVE_FORMAT_EXTENSIBLE` data) is that the lack of a channel mask and the particular GUIDs it uses unquestionably identify it as B-Format/FuMa audio, so it can't be mistaken.

The issue is that these audio loading APIs don't have a proper means for reporting that the loaded audio is B-Format. In some cases, they'll even ignore it when the file is identifies itself as such (e.g. ffmpeg seems to ignore unknown GUIDs, and a replaces a blank channel mask with a default based on channel count, so what's identifying itself as a 3D B-Format audio file is being reported as plain quad audio).

In addition to `amb` files, Ogg Opus recently gained support for B-Format (using AmbiX layout/scaling instead of FuMa). But in similar fashion, even though ffmpeg will load these files, it doesn't clearly report the audio as B-Format so an application can only guess based on the channel count and whether the channel mask is blank or not (unlike `amb` files, Opus B-Format gets reported with a blank channel mask). Worse, Opus files can contain extra data that's necessary for converting the stored audio to AmbiX layout/scaling, but that's completely ignored so files that rely on it will play broken regardless.

Ultimately, proper support for B-Format needs an audio loader that's capable of properly identifying and reporting audio as B-Format, which SDL_sound and ffmpeg currently don't. It also has to ensure a common layout and scaling, or at least report the layout and scaling, so the app can correctly interpret the audio.

PR #223, opened 2018-09-08
Fix SIGBUS for Android platform.

halx99:


kcat:
I'd rather find the real bug. The `src` pointer should be aligned, any number of other issues could arise if it's not (to say nothing of the performance impact of a 2- or 4-byte `memcpy` for each sample).

halx99:
Yes, you are right.

Issue #214, opened 2018-08-12
API extension to query and set period size and quantity


ceyt:
I've recently been bitten by a change in default number of periods, specifically, when GZDoom switched from 1.17.2 to more recent builds of OpenAL-soft:
https://forum.zdoom.org/viewtopic.php?f=50&t=61642

Similar problem was reported in issue #174. My sound card is ESI UGM96, the other report lists SB Audigy SE, my guess is that type of hardware reports very small buffer sizes that are suitable for optimized audio tools, but stress common games a bit too much. I'm not at the same machine at the moment, so I can't prove it with debug output, though.

While fixing the issue (by setting the value in the configuration file) is easy for any skilled user, it is not as straightforward from an application point of view. If I am not mistaken, there is no way for it to learn the number and length chosen for periods at initialization if it wants to warn the user, for example, and also no way to provide the desired values directly. Parsing and messing with the shared configuration file is obviously a wrong solution. As a result, application can neither display the active settings to the user, nor pass the user-provided values to OpenAL-soft back-end, unless it forks the library to add such functionality.

I am sure that GZDoom is not the only program that would benefit from having such interface.

kcat:
Can you get a trace log from a run using the default options? The original period count (4) with the default buffer size (1024 samples) and sample rate (44100hz) was quite high, and OpenAL Soft should already auto-scale up by necessary multiples if the system tries to impose different default values. For example, Windows by default sets a period size of 10ms, or 1/100th the sample rate, whereas OpenAL Soft defaults to 1024 samples at 44100hz, or ~21ms. OpenAL Soft then effectively doubles the period size for it to be 20ms, or 1/50th the sample rate. Similarly, if the device's sample rate is different than the default, the period count will auto-adjust to maintain the same buffer time.

I might add an extension to request the period size and count, but even if I do, it sounds like something is going wrong here that needs to be addressed. A system I have to test on is quite old (about 10 years), and is using an audio system that's known for not being very efficient (PulseAudio), and has no real trouble keeping up.

PR #227, opened 2018-09-14
Resolve conflict between SDL2 / OpenAL on Android

jgranick:
SDL2 and OpenAL both define `JNI_OnLoad`, leading to an error while linking.

This change causes OpenAL to rely upon SDL2 for gaining a reference to the current `JavaVM` environment when it is present

### DIFF
--- a/Alc/ALc.c
+++ b/Alc/ALc.c
@@ -1167,6 +1167,9 @@ static void alc_initconfig(void)
 
 #ifdef __ANDROID__
 #include <jni.h>
+#ifdef HAVE_SDL2
+#include <SDL.h>
+#endif
 
 static JavaVM *gJavaVM;
 static pthread_key_t gJVMThreadKey;
@@ -1178,6 +1181,9 @@ static void CleanupJNIEnv(void* UNUSED(ptr))
 
 void *Android_GetJNIEnv(void)
 {
+    #ifdef HAVE_SDL2
+    return SDL_AndroidGetJNIEnv();
+    #else
     if(!gJavaVM)
     {
         WARN("gJavaVM is NULL!\n");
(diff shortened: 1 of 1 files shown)


kcat:
I'm thinking the whole JNIEnv handling needs to be rethought. As far as I can tell, to get information like the device-native sample rate and update size (for low-latency processing) it's not enough to just have the `JNIEnv*`, there's also another Java object handle the app has that's needed. Without it, the JNI stuff is unnecessary since it's not used for anything else.

The two possible ways I see to handle it is to have an Android-only export/extension where an app can pass in the `JNIEnv*` and Java object handle. Though honestly at this point I remember so little of the JNI API that I don't know if I could get it to work.

The other possibility would be to implement an extension to request the playback buffer metrics, like suggested in issue #214. The application can already request the sample rate using the `ALC_FREQUENCY` content attribute, so also being able to set the update size would mean the app  could request the appropriate values via context creation attributes to achieve low-latency processing. Then there's no need for OpenAL Soft to worry about the JNIEnv at all. Though that would mean the app is responsible for getting that information from Java/JNI to give to OpenAL.

jgranick:
What is the other Java object handle we need to get this information?

kcat:
I believe it was the `android/content/Context` object, for the `getSystemService` method call. The Java code for getting the native sample rate and buffer/period size (which I need to call that through JNI, since there's no native C interface to get the information) is show here:

https://developer.android.com/ndk/guides/audio/audio-latency#output-latency

As I understand it, the `getSystemService` method is non-static, so it needs a `Context` object to make the call and retrieve the Audio Manager. Every Java-based Android app has one Context, given to them by the system, I believe. But there's no global static method to retrieve it, so the app has to provide it to whatever needs it.

kcat:
This shouldn't be an issue anymore with commit c39eeb963895a2eb05e17bc28c357f7d3f4ce815. Since OpenAL Soft doesn't actually use the `JNIEnv` for the time being, it's not necessary to have functions to get it.

PR #241, opened 2018-10-30
alcResetDeviceSOFT: try reactivating the device if it has been disconnected.

xxxbxxx:
If the device is disconnected, as detected using:
`alcGetIntegerv(m_Device, ALC_CONNECTED, 1, &Connected); `

the application can call `alcResetDeviceSOFT()`: it will try and succeed if the device has been reconnected.
The application then needs to reset the sources state as desired (by calling `alSourcePlay()`, ...).

### DIFF
--- a/Alc/ALc.c
+++ b/Alc/ALc.c
@@ -2366,7 +2366,7 @@ static void InitDevice(ALCdevice *device, enum DeviceType type)
     ALsizei i;
 
     InitRef(&device->ref, 1);
-    ATOMIC_INIT(&device->Connected, ALC_TRUE);
+    ATOMIC_INIT(&device->Connected, DeviceConnect_Connected);
     device->Type = type;
     ATOMIC_INIT(&device->LastError, ALC_NO_ERROR);
 
@@ -3257,7 +3257,7 @@ static ALCsizei GetIntegerv(ALCdevice *device, ALCenum param, ALCsizei size, ALC
                 values[i++] = ALC_CAPTURE_SAMPLES;
                 values[i++] = V0(device->Backend,availableSamples)();
                 values[i++] = ALC_CONNECTED;
-                values[i++] = ATOMIC_LOAD(&device->Connected, almemory_order_relaxed);
+                values[i++] = (ATOMIC_LOAD(&device->Connected, almemory_order_acquire) == DeviceConnect_Connected);
                 almtx_unlock(&device->BackendLock);
 
                 values[i++] = 0;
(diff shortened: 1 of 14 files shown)


Alc/ALc.c
```diff
@@ -4669,20 +4669,33 @@ ALC_API const ALCchar* ALC_APIENTRY alcGetStringiSOFT(ALCdevice *device, ALCenum
...
+        // try reconnect..
+        Connected = DeviceConnect_Connecting;
+        ATOMIC_STORE(&device->Connected, DeviceConnect_Connecting, almemory_order_relaxed);
+    }
 
     err = UpdateDeviceParams(device, attribs);
+
+    if (err == ALC_NO_ERROR)
+    {
+        ATOMIC_COMPARE_EXCHANGE_WEAK(&device->Connected, &Connected, DeviceConnect_Connected, almemory_order_acq_rel, almemory_order_acquire);
```
kcat: This doesn't look quite right. This almost certainly should be `_STRONG` (weak cmpxchg is allowed to spuriously fail and not swap, even if the comparison is true). There's also a bit of a race condition since you changed the backends' mixing loops to check for `DeviceConnect_Connected`, so if the backend loop happens to run and see `DeviceConnect_Connecting` before this gets a chance to change it, it'll stop mixing despite having successfully restarted.

kcat:
It may be simpler to just say that when `alcResetDeviceSOFT` is called on a disconnected device, the device's connection status is first restored before resetting/restarting playback (and should resetting/playback fail, it will logically become disconnected again). While it's true that most OpenAL calls are atomic (any observable state as a result of a call will agree that the call has not (yet) executed, or has completed successfully; no partially-executed state can be seen), there a few exceptions to that rule. And since behavior between the `alcResetDeviceSOFT` call and the `ALC_EXT_DISCONNECT` extension is undefined (that undefined behavior being what allows us to implement reconnection in the first place), it might not be too problematic to be a bit lax here. That is, unless there is a desire to be atomic.

In either case, it still needs to be careful of race conditions, if there's any code that could behave badly if the device becomes connected again after having been disconnected, or a function misbehaving with the connected status switching from disconnected to connected to disconnected again in quick succession. I can't think of anything off the top of my head.

kcat:
Commit 598851fed7dfdd787de3487002afb13ed6e9dc82 now attempts to reconnect a lost device. It does so by setting the `Connected` flag back to true before resetting the backend device. If there's anything more a backend (e.g. WASAPI) should do to recover the same device it lost, it should do so during the backend device's `reset` method (or `resetProxy` for WASAPI).

kcat:
I'm closing this since the general behavior is implemented, and this can't apply to the current code base anymore anyway. If there's anything more specific that's desired, feel free to create a new PR.

PR #244, opened 2018-11-02
Cleanup alBuffer.cpp

ShFil119:
Used c++ libs, changed argument name to match with implementation.

### DIFF
--- a/OpenAL32/alBuffer.cpp
+++ b/OpenAL32/alBuffer.cpp
@@ -20,10 +20,11 @@
 
 #include "config.h"
 
-#include <stdlib.h>
-#include <stdio.h>
-#include <assert.h>
-#include <limits.h>
+#include <cassert>
+#include <climits>
+#include <cstdio>
+#include <cstdlib>
+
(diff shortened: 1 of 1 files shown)


OpenAL32/alBuffer.cpp
```diff
@@ -20,10 +20,11 @@
...
 #include "config.h"
 
-#include <stdlib.h>
-#include <stdio.h>
-#include <assert.h>
-#include <limits.h>
+#include <cassert>
+#include <climits>
+#include <cstdio>
+#include <cstdlib>
```
kcat: Won't these require using the `std::` namespace for the functions and types, e.g. `std::memcpy`?

ShFil119: Yeah. Some implementation also have function in global namespace and std. I'll recheck usages. 

Reason for change is that `c` libs are less liked by templates.

ShFil119:
I guess (as conversion to c++ is in progress) this kind of cleanup is useless for now.

PR #272, opened 2019-02-18
Remove some possible sources of crash

ShFil119:
It's possible to catch these scenarios (nullptr deference / dividing by zero) using clang-analyzer.

### DIFF
--- a/Alc/backends/null.cpp
+++ b/Alc/backends/null.cpp
@@ -96,7 +96,7 @@ int NullBackend::mixerProc()
          * and current time from growing too large, while maintaining the
          * correct number of samples to render.
          */
-        if(done >= mDevice->Frequency)
+        if(done >= mDevice->Frequency && mDevice->Frequency != 0)
         {
             seconds s{done/mDevice->Frequency};
             start += s;
--- a/OpenAL32/alSource.cpp
+++ b/OpenAL32/alSource.cpp
@@ -54,6 +54,9 @@ using namespace std::placeholders;
 
 inline ALvoice *GetSourceVoice(ALsource *source, ALCcontext *context)
 {
+    if(!source || !context)
+        return nullptr;
+
     ALint idx{source->VoiceIdx};
     if(idx >= 0 && idx < context->VoiceCount.load(std::memory_order_relaxed))
     {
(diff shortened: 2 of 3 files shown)


Alc/backends/null.cpp
```diff
@@ -96,7 +96,7 @@ int NullBackend::mixerProc()
          * and current time from growing too large, while maintaining the
          * correct number of samples to render.
          */
-        if(done >= mDevice->Frequency)
+        if(done >= mDevice->Frequency && mDevice->Frequency != 0)
```
kcat: It's impossible for `Frequency` to be 0 (it's initialized to 44100 by default, and the minimum allowed is 8000). If it's 0, something else is very wrong and many other problems will occur. Perhaps an `ASSUME(device-Frequency > 0);` statement can silence clang-analyzer?

OpenAL32/alSource.cpp
```diff
@@ -54,6 +54,9 @@ using namespace std::placeholders;
 
 inline ALvoice *GetSourceVoice(ALsource *source, ALCcontext *context)
 {
+    if(!source || !context)
```
kcat: These can't be null, since they should already have been checked (if not used) before the call. If they manage to be null is some situation, it needs to be detected earlier so the appropriate error can be set before getting to this point.

OpenAL32/alSource.cpp
```diff
@@ -2703,9 +2706,12 @@ AL_API ALvoid AL_APIENTRY alSourcePlayv(ALsizei n, const ALuint *sources)
             [&context](ALuint sid) -> void
             {
                 ALsource *source{LookupSource(context.get(), sid)};
-                source->OffsetType = AL_NONE;
-                source->Offset = 0.0;
-                source->state = AL_STOPPED;
+                if(source)
```
kcat: These also can't be null since the IDs were looked up earlier in the call and verified to be non-null (the source lock is also being held so a call on another thread can't delete the ID in between). If it manages to be null, an error needs to be set before any modification happens to the sources.

OpenAL32/alSource.cpp
```diff
@@ -2723,6 +2729,9 @@ AL_API ALvoid AL_APIENTRY alSourcePlayv(ALsizei n, const ALuint *sources)
     auto start_source = [&context,device](ALuint sid) -> void
     {
         ALsource *source{LookupSource(context.get(), sid)};
+        if(!source)
+            return;
```
kcat: Same here. It was checked earlier under the same lock, and if it gets a null here there's bigger problems.

ShFil119:
False positives, so closing.

PR #285, opened 2019-04-29
Update CMakeLists.txt. Simplify examples build and export PUBLIC headers.

ArthurSonzogni:
Hi,

These two changes are ready for reviews. I am mostly interested in the second one "export PUBLIC header". The first one is a bit more opinionated, please let me know if you prefer to skip it.


utils/alsoft-config/CMakeLists.txt
```diff
@@ -56,3 +56,4 @@ else()
         )
     endif()
 endif()
+target_include_directories(alsoft-config PRIVATE "${OpenAL_BINARY_DIR}")
```
kcat: This seems to break the build if `alsoft-config` isn't getting built. It should probably be in a `if(TARGET alsoft-cobfig)` check, or put it in the two blocks that declare the target (one for Qt5 and another for Qt4).

CMakeLists.txt
```diff
@@ -1539,68 +1540,38 @@ IF(ALSOFT_EXAMPLES)
...
+                PUBLIC
+                    ${SDL2_INCLUDE_DIR}
+                    ${SDL_SOUND_INCLUDE_DIR}
+            )
+            TARGET_LINK_LIBRARIES(ex-common
+                PUBLIC
+                    ${SDL_SOUND_LIBRARIES}
+                    ${SDL2_LIBRARY}
+                    ${MATH_LIB}
+            )
```
kcat: Does this work as seemingly intended, that it only applies to the future uses of `ex-common` and not retroactively?

utils/alsoft-config/CMakeLists.txt
ArthurSonzogni: Thanks!
I think I will just duplicate the line in both cases, like it is done very everything else.
Ideally, it will have a patch to deduplicate everything here in the future.

CMakeLists.txt
ArthurSonzogni: No, it applies it retroactively AFAIK. It means the example "alrecord" will linked with the SDL unnecessary.

These are just examples, so I guessed it  isn't really an issue. If it is an issue, I can skip this patch. What do you think?

kcat: You could probably instead create a new target (e.g. `ex-common-sdl`) that includes ex-common and those additional SDL flags, then use that for the examples that need SDL. Otherwise, it would be better to skip the patch so examples don't link against SDL and ffmpeg unnecessarily.

CMakeLists.txt
```diff
@@ -1314,6 +1316,15 @@ ELSE()
...
     ENDIF()
 ENDIF()
+
+TARGET_INCLUDE_DIRECTORIES(${IMPL_TARGET}
+ PUBLIC
+   $<BUILD_INTERFACE:${OpenAL_SOURCE_DIR}/include>
+   $<INSTALL_INTERFACE:include>
+ PRIVATE
+   "${OpenAL_BINARY_DIR}"
+)
```
kcat: It would probably make sense to combine this with the existing `TARGET_INCLUDE_DIRECTORIES(${IMPL_TARGET} ...` call a few lines down.

ArthurSonzogni: Okay. I will do it.

CMakeLists.txt
ArthurSonzogni: I skipped patch #1.

kcat:
The source/include, source/common, and binary paths need to be added to the router's include directories as well.

ArthurSonzogni:
Sorry, I'm having a little trouble today. I am a bit confused with the router and the OpenAL target being either the router or the implementation.

CMakeLists.txt
```diff
@@ -1309,22 +1321,36 @@ ELSE()
...
+    ADD_LIBRARY(${IMPL_TARGET} SHARED)
     IF(WIN32)
         SET_TARGET_PROPERTIES(${IMPL_TARGET} PROPERTIES PREFIX "")
     ENDIF()
 ENDIF()
+
+TARGET_SOURCES(${IMPL_TARGET} PRIVATE ${OPENAL_OBJS} ${ALC_OBJS})
+
+TARGET_LINK_LIBRARIES(${IMPL_TARGET}
+    PRIVATE common ${LINKER_FLAGS} ${EXTRA_LIBS} ${MATH_LIB})
```
kcat: Unfortunately you can't link to `common` like this when the openal implementation is static. A static library can't link to another library. At best, CMake will track a static library target's dependencies and automatically include them as needed when linking, but the generated static `libopenal.a` library should be stand-alone and not need a separate `libcommon.a` with it.

ArthurSonzogni: Hmm, Sorry!
In this case, would it make sense just to abandon the target "common" and simply include the sources to the libraries?

kcat: Normally I'd prefer not to, but since it's only actually needed for these two libraries now, I suppose I can't see the harm.

CMakeLists.txt
```diff
@@ -1277,12 +1269,8 @@ IF(LIBTYPE STREQUAL "STATIC")
     IF(WIN32 AND ALSOFT_NO_UID_DEFS)
         SET(CPP_DEFS ${CPP_DEFS} AL_NO_UID_DEFS)
     ENDIF()
-    ADD_LIBRARY(OpenAL STATIC ${COMMON_OBJS} ${OPENAL_OBJS} ${ALC_OBJS})
+    ADD_LIBRARY(OpenAL STATIC ${COMMON_OBS} ${OPENAL_OBJS} ${ALC_OBJS})
```
kcat: Looks like a typo slipped in here.

ArthurSonzogni:
Thanks!

Issue #288, opened 2019-05-08
BackendListEnd not initialized.


ArthurSonzogni:
Please take a look at my terminal:
[![asciicast](https://asciinema.org/a/o7Sr4kiHx4vLdWA8ct9mMEl9D.svg)](https://asciinema.org/a/o7Sr4kiHx4vLdWA8ct9mMEl9D)

We can see that the global variable BackendListEnd is equal to 0x0.

It looks like it isn't initialized yet. Moving its declaration + initialization from the global scope to the function alc_config_init scope fixed the issue.

global initialization with DLL and threads might not work very well in C++.

I am working on a quick fix.

ArthurSonzogni:
I was initializing OpenAL from the constructor of a static object, which in my case, happened before openal initialize its static variables.

So, this is 95% my fault and I will close this issue. This could also be solved by removing every OpenAL static variable requiring a runtime initialization.

PR #289, opened 2019-05-08
Move BackendListEnd variable from global scope to function scope.

ArthurSonzogni:
It reduces the scope of this variable to the function it is used, which
is cleaner.

It also avoid using the variable before it has been initialized.
Cf  StackOverflow: https://stackoverflow.com/a/25009120/5112390
~~~
The standard says that all objects are initialized in the same
translation unit (aka object file which corresponds to a single source
file) before any function is called in that translation unit.

The standard is allowing lazy initialization to occur in the event that
a DLL is loaded at run time. If you allow run-time linkage of your code,
you can't say that everything is initialized before main().
~~~

Bug: https://github.com/kcat/openal-soft/issues/288

### DIFF
--- a/Alc/alc.cpp
+++ b/Alc/alc.cpp
@@ -178,7 +178,6 @@ BackendInfo BackendList[] = {
     { "wave", WaveBackendFactory::getFactory },
 #endif
 };
-auto BackendListEnd = std::end(BackendList);
 
 BackendInfo PlaybackBackend;
 BackendInfo CaptureBackend;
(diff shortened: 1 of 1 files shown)


kcat:
If it's the case that global variables may not be initialized before main, and a call is made that causes OpenAL Soft to initialize before main, I don't think this change would be enough. There are several other global and static member variables throughout the library, and if they're not initialized yet they would not only be seen with incorrect values, but the values they get set to would be overwritten with the defaults they're globally initialized to.

ArthurSonzogni:
Yes and moreover I am not sure my patch is correct. I guess you can't init
openal several times anymore. I am seeing this function is modifying the
global array.

I will try to minimize my reproducer to really understand what happens
specifically in my case.

ArthurSonzogni:
Yes, this was 100% my fault.

alc_initconfig was called from the initialization of one of my static variable, which happens before the initialization of openal-soft static variables.

It would be nice to remove most of the global variables from openal to make it more resilient to this kind of usage, but I guess, as you said, this change won't be enough.

Note: Previously, I was linking against libopenal-dev, and it was working. Now, I am linking against openal-soft shared library built by this cmake.

PR #290, opened 2019-05-12
Refactor BackendList / alc_initconfig / ALSOFT_DRIVERS.

ArthurSonzogni:
1. Avoid global variables. If users init OpenAL before main(), some
   variables with runtime initialization may not be initialized yet.
   1.1. Remove |BackendListEnd|.
   1.2. Move |BackendList| into alc_initconfig.

2. Refactor ALSOFT_DRIVERS parser + usage. The previous version was a bit
   complex to understand for newcomers.

### DIFF
--- a/Alc/alc.cpp
+++ b/Alc/alc.cpp
@@ -129,57 +129,6 @@ struct BackendInfo {
     BackendFactory& (*getFactory)(void);
 };
 
-BackendInfo BackendList[] = {
-#ifdef HAVE_JACK
-    { "jack", JackBackendFactory::getFactory },
-#endif
-#ifdef HAVE_PULSEAUDIO
-    { "pulse", PulseBackendFactory::getFactory },
-#endif
-#ifdef HAVE_ALSA
-    { "alsa", AlsaBackendFactory::getFactory },
-#endif
-#ifdef HAVE_WASAPI
-    { "wasapi", WasapiBackendFactory::getFactory },
-#endif
-#ifdef HAVE_COREAUDIO
-    { "core", CoreAudioBackendFactory::getFactory },
-#endif
-#ifdef HAVE_OPENSL
-    { "opensl", OSLBackendFactory::getFactory },
-#endif
-#ifdef HAVE_SOLARIS
-    { "solaris", SolarisBackendFactory::getFactory },
-#endif
-#ifdef HAVE_SNDIO
-    { "sndio", SndIOBackendFactory::getFactory },
-#endif
-#ifdef HAVE_OSS
-    { "oss", OSSBackendFactory::getFactory },
-#endif
-#ifdef HAVE_QSA
-    { "qsa", QSABackendFactory::getFactory },
-#endif
-#ifdef HAVE_DSOUND
-    { "dsound", DSoundBackendFactory::getFactory },
-#endif
-#ifdef HAVE_WINMM
-    { "winmm", WinMMBackendFactory::getFactory },
-#endif
-#ifdef HAVE_PORTAUDIO
-    { "port", PortBackendFactory::getFactory },
-#endif
-#ifdef HAVE_SDL2
-    { "sdl2", SDL2BackendFactory::getFactory },
-#endif
-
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Hi, kcat@,

What about this? Could you please take a look?

I haven't found ways to add tests, so I only wrote a few ones locally to check it works. Then I had to remove them.
Maybe we could write and run some tests. To avoid asking the developers to download and install a third_party library, one can use cmake[FetchContent](https://cmake.org/cmake/help/git-stage/module/FetchContent.html) to download gtest?

kcat:
There's more global variables that may be touched aside from just that backend list. The `std::once_flag` itself, the selected playback and capture backends' storage, alu.cpp, helpers.cpp, the various backends, and perhaps more. It's really not safe to try initializing OpenAL before global initialization is complete.

What are you doing before main that's causing OpenAL to initialize?

ArthurSonzogni:
Yes, I agree this is still not safe to use OpenAL-soft before main(). But I
guess we can remove every global variables that are: read before being
write AND requiring a runtime initialization.

I am interested supporting this, because I already have created a
multimedia library (like SFML) and a game using it. What I did is putting
my two objects storing and playing sound behind a refcount that will
initialize OpenAL when required. These objects (even when not
loading/playing sounds) will increment the refcount and initialize OpenAL.

With the emscripten/OpenAL implementation, it worked. More strange, when I
was using the package libopenal-dev, it was also working. When I switched
to using OpenAL from source, it started to fail, only because of the
variable backenlistend.

Even if we are still far from supporting this kind of usage, I think my
patch doesn't hurt. That's why I phrased it like a refactor an wasn't
claiming it was fixing a bug.

kcat:
> But I guess we can remove every global variables that are: read before being write AND requiring a runtime initialization.

A big problem there is knowing which variables that applies to. C++ has certain rules on which global or static variables can effectively be set at compile time (i.e. like C, where a variable has a value set in the compiled data storage), and which actually needs to run code when starting the app to generate and set the value. You can't really tell which it is just from looking at the variable's initialization.

> These objects (even when not loading/playing sounds) will increment the refcount and initialize OpenAL.

Initialize in what way? Currently, there's only two things that will cause OpenAL to do this initialization: opening a device (playback, capture, or loopback), or enumerating devices. This is even less safe before main, because that involves OpenAL doing many more things before being fully initialized (including potentially creating and messaging back-and-forth with threads, IPC, and getting interrupts set up), and also depends on other libraries and system services working correctly.

> With the emscripten/OpenAL implementation, it worked. More strange, when I was using the package libopenal-dev, it was also working. When I switched to using OpenAL from source, it started to fail, only because of the variable backenlistend.

I can't speak about the emscripten/OpenAL implementation, but the libopenal-dev package is likely for OpenAL Soft 1.19.1 or earlier, which is C. As mentioned before, C's global initialization essentially requires knowing values at compile-time, and thus not depend on run-time initialization. GCC/MSVC extensions were used to do a little bit of run-time initialization on the few things that needed it, but it was less depended on and likely this behaved slightly differently than C++ global initialization.

Either way though, doing threading or starting audio processing prior to main would be a big no-no, and that it worked before I'd consider to be pure luck. Even if I managed to get OpenAL Soft to avoid any problems there, I can't make any guarantee about the libraries and services actually used for audio I/O. I can't tell you if Windows would be happy setting up COM and doing IPC prior to main, nor can I tell you if libpulse may one day start using C++ and cause a similar issue on its end. Perhaps backend X will work but not backend Y, and/or perhaps it'll work 99% of the time but fail randomly 1% of the time for no apparent reason.

ArthurSonzogni:
Yes, it makes sense. You can provide this guarantee, because your dependencies don't.

In itself, my commit doesn't hurt and I think it makes the code a little bit more readable . Let me know if there are parts you like or not. Otherwise, feel free to close this pull request ;-)

kcat:
This is currently failing for MSVC, and it changes the behavior of the name list (a list that doesn't end in `,` shouldn't contain unspecified backends).

Issue #287, opened 2019-05-04
Xcompiling fails with "error: a reinterpret_cast is not a constant expression"


DeadSix27:
Hey, when trying to cross-compile openal to win64 I get the following error:

```
/Alc/alc.cpp:190:23: error: a reinterpret_cast is not a constant expression
  190 | #define DECL(x) { #x, (ALCvoid*)(x) }
```
Is it possible that using different gcc versions for native and cross compiling causes this?
```
$ x86_64-w64-mingw32-gcc --version
x86_64-w64-mingw32-gcc (GCC) 9.1.0

$ gcc --version
gcc (GCC) 8.3.0
```

Logs:
native_tools_CMakeOutput.log https://0x0.st/zc5G.log
CMakeError.log https://0x0.st/zc5D.log
CMakeOutput.log https://0x0.st/zc5k.log

<details>
<summary>make output</summary>

```bash
bash-5.0$ make -j4
[  1%] Built target native-tools
[  1%] Built target build_version
[  3%] Building CXX object CMakeFiles/OpenAL.dir/Alc/alc.cpp.obj
[  4%] Building CXX object CMakeFiles/OpenAL.dir/Alc/effects/distortion.cpp.obj
[  6%] Building CXX object CMakeFiles/OpenAL.dir/Alc/effects/dedicated.cpp.obj
[  7%] Building CXX object CMakeFiles/OpenAL.dir/Alc/effects/echo.cpp.obj
/xc/workdir/x86_64/openal-soft_git/Alc/alc.cpp:190:23: error: a reinterpret_cast is not a constant expression
  190 | #define DECL(x) { #x, (ALCvoid*)(x) }
      |                       ^~~~~~~~~~~~~
/xc/workdir/x86_64/openal-soft_git/Alc/alc.cpp:195:5: note: in expansion of macro 'DECL'
  195 |     DECL(alcCreateContext),
      |     ^~~~
In file included from /xc/workdir/x86_64/openal-soft_git/OpenAL32/Include/alu.h:20,
                 from /xc/workdir/x86_64/openal-soft_git/OpenAL32/Include/alSource.h:7,
                 from /xc/workdir/x86_64/openal-soft_git/Alc/alc.cpp:43:
/xc/workdir/x86_64/openal-soft_git/Alc/alc.cpp: In function 'ALCenum UpdateDeviceParams(ALCdevice*, const ALCint*)':
[...]

PR #301, opened 2019-06-28
Fix the error "a reinterpret_cast is not a constant expression" in GCC 9

gongminmin:
reinterpret_cast and C-style casting is no longer a constexpr in GCC9. Need to fall back to const instead.


kcat:
Is `reinterpret_cast` no longer `constexpr` in C++, or is that a GCC bug? It was brought up in issue #287, but haven't heard back whether it's a compiler bug or not.

gongminmin:
It's in C++11 standard. Gcc supports reinterpret_cast as constexpr is a bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=49171. And it's fixed in 9. That's why this problem pops out now.

PR #309, opened 2019-07-02
Add noexcept to match the signature of sampler functions

gongminmin:
This change fixes compilation problems on gcc 9.1 with -std=c++17.

### DIFF
--- a/Alc/mixer/mixer_c.cpp
+++ b/Alc/mixer/mixer_c.cpp
@@ -14,13 +14,13 @@
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


kcat:
C++11/14 doesn't allow `noexcept` on types. The exception specification didn't become part of the function type until C++17.

I don't see why this would be a problem, though. A `noexcept`/`noexcept(true)` function can be assigned to (and called though) a function pointer/reference with `noexcept(false)` (or without `noexcept`) -- the caller simply assumes it could throw, but it won't so it's doesn't affect anything.

gongminmin:
Thanks!

PR #312, opened 2019-07-10
vocal morpher implementation

Lopuska:
Hi! : )
Here we are finally.
The best way to test it, is by using a white/pink noise (as it is supposed to be used in a production use case)
Changes are minimal and code does just what is necessary.
Pitch shifting is not in yet, I wanted to push it in a separate PR.


kcat:
One of these days, I should get around to making an example or utility that can set up a user-defined effect and its parameters to play a sound through to test these effects, instead of relying on the SDK's EFX10Show in Wine or something.

Lopuska:
@kcat Uh nice! If you think that might be useful, I can open source the app I made back in time:
image
You could even include it in the repo.
Unfortunately it's a MFC windows only executable but very useful when testing DSP. Maybe a port to Qt is something I should consider.

PR #317, opened 2019-07-16
misc fixes and improvements for Vocal Morpher

Lopuska:
Added few comments, unify some code style to be consistent with the others oal soft effects.
Fixed a logic error during filter params copy.
The part which includes pitch shifter is ready and I'll soon send the PR but I don't like the fact that I do a lot of boilerplate because of code duplication. If for you is fine I can unify the code between pshifter.cpp and vmorpher.cpp to avoid code duplication. My idea is to add a utility function in al_complex.h and a structure which represent all the buffers for the FFT pitch shifting. That will allow much easily to perform pitch shifting around the codebase, in just few lines of code.

### DIFF
--- a/Alc/effects/vmorpher.cpp
+++ b/Alc/effects/vmorpher.cpp
@@ -22,7 +22,6 @@
 
 #include <cmath>
 #include <cstdlib>
-
 #include <algorithm>
 #include <functional>
 
@@ -32,7 +31,6 @@
 #include "alError.h"
 #include "alu.h"
 #include "vecmat.h"
-
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


kcat:
> The part which includes pitch shifter is ready and I'll soon send the PR but I don't like the fact that I do a lot of boilerplate because of code duplication. If for you is fine I can unify the code between pshifter.cpp and vmorpher.cpp to avoid code duplication.

I thought the pitch shift for the vocal morpher was changing the formant frequencies, not the input sound's frequencies? Given that the vocal morpher knocks out all but four set frequency bands, and the input is generally supposed to be wideband noise, shifting the input sound doesn't seem very useful. Shifting the format frequencies, however, would effectively change the apparent vocal pitch.

Lopuska:
that's right, then I'll follow that way, it's even simpler.

PR #318, opened 2019-07-18
pitch shift for formant filters

Lopuska:
Final part done! : D


kcat:
Hm, well there's the other phonemes too (aa, ae, etc) ;). We still also need to figure out what's supposed to be done with the consonant sounds.

Lopuska:
That's right.
For the blended vowels, is going to be easy adding them.
For the consonants, we need to find a reference on how they are supposed to work.

Issue #325, opened 2019-08-14
[Suggestion] On demand source buffer?


shangjiaxuan:
Is there any plans to implement a software synthesizer when OpenAL needs a source? 

Similar to the current use case of using a loopback device to implement custom backend, maybe there can be a extension that registers a function to be called when the data is needed? (Of course, also need some non-block guarantee here.) 

This can reduce the resource usage when playing compressed long audio without the need for extra memory copying and usage (e.g. plugging a decoder via a user pointer, and requesting a number of frames on the interface) compared to buffer queuing (and also more flexible if the user want a dynamic software synthesizer as source (also less latency when user wants to change the buffer contents)).

Maybe some interface like this to bind a generator to a buffer?
`void alBufferFuncSOFT(ALuint abo, ALenum format, ALsizei freq, alSourceFunc_t func, ALvoid* usr_ptr)`
`typedef Alsizei(*alSourceFunc_t)(Aluint abo, ALvoid* to_fill, ALuint min_frames, ALuint max_frames, ALvoid* usr_ptr /*, ALuint alignment, Alenum format, ...*/)`

(There may need to be format conversion set up after each call to callback, so maybe setting up format beforehand or providing more parameters to the callback and avoiding this conversion may be better but will require more effort on the user side).

The format, sample rate, etc can still be accessed the `alBufferxx`, `alGetBufferxx` interface, and buffer queuing and other things will still work, since it's still basically just a buffer on demand.

(If there is any plan on implementing this, I'm happy to dig down to the source and help...)

shangjiaxuan:
Also, currently the `BackendBase::mMutex` is only called by the `BackendBase::lock()` virtual method, which in turn is only called in every buffer write callback (and a few functions to retrieve backend status). It is possible to remove the need to lock it entirely?

kcat:
> This can reduce the resource usage when playing compressed long audio without the need for extra memory copying and usage (e.g. plugging a decoder via a user pointer, and requesting a number of frames on the interface)

This would actually be trickier than it may seem. Unless you know the exact behavior of a given decoder, you can't be sure that the decode function won't employ some locks (to ensure safe multi-threaded access to internal state) or allocations (for on-demand internal storage). With callbacks invoked by the mixer, you shouldn't be doing these things as they can stall processing without warning and make the audio update run late. Made worse by the fact that some backends may be more sensitive to this kind of thing than others, so you may not see a problem until a user with a particular system or configuration tries it.

[...]

PR #336, opened 2019-09-01
Adding a callback buffer type

shangjiaxuan:
#325 I've made a few changes to make this functionality. Looked up the buffer APIs and added a few checks for incompatible calls. The Gen and Delete functions for the buffer objects currently uses placement new and delete, so as long as the constructor initializes the function pointers to null, the buffer will be in a valid state.

If there's problems with this, I will look into it later. Thanks!


kcat:
It doesn't look like there's any changes here. Just a merge commit with nothing in it.

shangjiaxuan:
I'm really sorry that I forgot to merge back the working branch. I will also clean up the naming conventions right away...

al/source.cpp
```diff
@@ -3301,8 +3304,12 @@ ALsource::~ALsource()
...
         std::unique_ptr<ALbufferlistitem> head{BufferList};
-        BufferList = head->mNext.load(std::memory_order_relaxed);
-        if(ALbuffer *buffer{head->mBuffer}) DecrementRef(buffer->ref);
+		/* If (BufferList->next==nullptr), the following will segfault, 
+		* thus the loading of next is moved to the next line, hopefully
+		* this won't break anything.
+		*/
+        if(ALbuffer *buffer{BufferList->mBuffer}) DecrementRef(buffer->ref);
+		BufferList = head->mNext.load(std::memory_order_relaxed);
     }
```
shangjiaxuan: This is not needed with  the `if(ALbuffer *buffer{head->mBuffer}) DecrementRef(buffer->ref);`

al/source.cpp
```diff
@@ -3301,8 +3304,12 @@ ALsource::~ALsource()
     while(BufferList != nullptr)
     {
         std::unique_ptr<ALbufferlistitem> head{BufferList};
-        BufferList = head->mNext.load(std::memory_order_relaxed);
-        if(ALbuffer *buffer{head->mBuffer}) DecrementRef(buffer->ref);
+		/* If (BufferList->next==nullptr), the following will segfault, 
+		* thus the loading of next is moved to the next line, hopefully
+		* this won't break anything.
+		*/
+        if(ALbuffer *buffer{BufferList->mBuffer}) DecrementRef(buffer->ref);
```
shangjiaxuan: This is not needed with  the `if(ALbuffer *buffer{head->mBuffer}) DecrementRef(buffer->ref);`

kcat:
The main problem I see is that this won't work with non-mono formats since it's running the callback in the per-channel loop (so each channel of the format will invoke the callback separately). The callback will need to be invoked after the number of source samples is determined, before working on each channel.

There would also need to be some protection that the buffer can't be queued onto a source, and probably can't be set on multiple sources at once.

Also I'm not yet sure the best way to deal with the resampler padding. Always using all the stored samples would essentially create a constant delay, which would cause a problem for apps that want synchronization between callback and non-callback sources. I think what would be needed here is some kind of "first mix" flag that indicates it needs it get extra samples from the callback to fill in the resampler padding, and then with subsequent mixes it can restore all the padding samples and only get new samples from the callback (the `VOICE_IS_FADING` flag kind of already serves this purpose, in reverse).

It might also be a good idea to make it so when the callback provides less than the requested number of samples, it makes the source stop.

shangjiaxuan:
A thought on stereo output is to use planar output for stereo or more channels (more arguments), and moving the buffer load of callback outside the loop seems like a valid solution, but that would require more intrusive modifications.

You are right about stopping the source when buffer ends is a good idea, I will look into that. For the re-sampler padding thing I think it's just a matter of offset in the load. I think it may be implemented by calling the callback as soon as it is created to restore the offset.

kcat:
> A thought on stereo output is to use planar output for stereo or more channels (more arguments)

I'm not sure that's a good idea since currently all buffer data is expected to be interleaved channels. Planar layouts for multi-channel formats would be a good idea, but it would be specified by the caller ahead of time (likely as buffer properties, allowing the app to more generally handle padding between channels) and be independent of whether it's a callback or not.

> But still, you will need to know the following samples to actually do the re-sampling. Thus it will always be off by the padding to the future, and it seems a flag cannot really fix the inherent need to know some samples ahead.

In a sense, that's true. The callback will always have sent samples ahead of the current playback position. But the idea is that a source with a normal buffer and a source with a callback buffer can be started at the same time (with `alSourcePlayv`), and if the callback provides the exact same sample data the normal buffer has, the two should be heard perfectly synchronized.

> Also the audio thread works on a time resolution of each callback from the underlying back-end. Any time that is smaller is supposed to be dealt with implicitly (aka, works like a buffer stream with the exact buffer size needed, with no queuing and mapped memory).

I'm not sure what you mean. But as far as the backend is concerned, the update size isn't fixed. Some backends, like PulseAudio and WASAPI, will mix however many samples it can when it's told to. But even other backends, like ALSA or DSound, will mix a multiple of the period size if there's enough space to. It should not try to mix less than one update/period size, that much is guaranteed, but it can mix anything equal to or greater than one update/period. Plus, given the sample rate of the buffer and any pitch shifting on the source, the mixer may need to break up a mix into multiple partial mixes so it doesn't overflow certain internal buffers.

kcat:
Apologies for this, but I got inspired yesterday and worked out an implementation that should work correctly with synchronization, multi-channel formats, and stopping at the end. I haven't gotten around to testing it yet, and the extension isn't finalized, so if you have any suggestions for changes or you want to get your example working with it, you're more than welcome.

shangjiaxuan:
I will look at this now. My test is not really comprehensive, though.

PR #343, opened 2019-09-15
remove few redundant temp var copy

Lopuska:


alc/effects/autowah.cpp
```diff
@@ -128,31 +128,22 @@ void ALautowahState::update(const ALCcontext *context, const ALeffectslot *slot,
...
 void ALautowahState::process(const size_t samplesToDo, const al::span<const FloatBufferLine> samplesIn, const al::span<FloatBufferLine> samplesOut)
 {
-    const ALfloat attack_rate = mAttackRate;
-    const ALfloat release_rate = mReleaseRate;
-    const ALfloat res_gain = mResonanceGain;
-    const ALfloat peak_gain = mPeakGain;
-    const ALfloat freq_min = mFreqMinNorm;
-    const ALfloat bandwidth = mBandwidthNorm;
-
-    ALfloat env_delay{mEnvDelay};
```
kcat: These copies are actually useful for optimizations. The compiler must assume a pointer or reference to a float can alias other non-local floats (same for any other matching type), so after it writes a float through a pointer/reference, it must reload any other non-local floats in case it was the same one. By having a local const copy, the compiler can know a reference it writes to can't alias that copy (a function can't be given a pointer/reference to an object that didn't yet exist), so it's free to keep it in a register after writing.

There are some cases where it doesn't matter to much. For example, after a function call if it needs to reload a variable from the stack, reloading it from `this` probably isn't too different (cache pressure not withstanding). Or if it only writes to something it knows can't alias the member variables, though this seems trickier to prove for the compiler.

Lopuska: fair enough, good catch!

PR #355, opened 2019-11-07
EFX: Reverb modulation

Raulshc:
I've reimplemented the modulation stage using the original code of Christopher Fitzgerald. The code was essentially correct and should be applied only to the feedback taps of late reverberation.

The original code is taken from v1.17.

Please, review and merge it at yout convenience (before or after 1.20 release).

### DIFF
--- a/alc/effects/reverb.cpp
+++ b/alc/effects/reverb.cpp
@@ -1426,7 +1426,7 @@ ALfloat ReverbState::calcModulation(ALfloat sample, size_t offset, size_t c)
      * sampling rate).  The center of the sinus is moved to reduce the delay
      * of the effect when the time or depth are low. 
      */
-    sinus = 1.0f - cosf(al::MathDefs<float>::Tau() * mMod.Index[c] / mMod.Range[c]);
+    sinus = 1.0f - std::cos(al::MathDefs<float>::Tau() * mMod.Index[c] / mMod.Range[c]);
 
     /* Step the modulation index forward, keeping it bound to its range. */
     mMod.Index[c] = (mMod.Index[c]+1) % mMod.Range[c];
--- a/alc/effects/reverb.cpp
+++ b/alc/effects/reverb.cpp
@@ -60,6 +60,21 @@ constexpr size_t MAX_UPDATE_SAMPLES{256};
  */
 constexpr size_t NUM_LINES{4u};
 
+/* This coefficient is used to define the maximum frequency range controlled
+* by the modulation depth.  The current value of 0.05 will allow it to swing
+* from 0.95x to 1.05x.  This value must be below 1.  At 1 it will cause the
+* sampler to stall on the downswing, and above 1 it will cause it to sample
+* backwards. The value 0.05 seems be nearest to Creative hardware behavior.
+*/
+constexpr ALfloat MODULATION_DEPTH_COEFF{0.05f};
+
+/* A filter is used to avoid the terrible distortion caused by changing
+* modulation time and/or depth.  To be consistent across different sample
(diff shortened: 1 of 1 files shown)


Raulshc:
BTW, is the reverb echo stage implemented in the current code?

kcat:
Are there any examples of what it sounds like on real hardware, or any other wide-spread reverb implementation? I removed this code because I had serious doubts that it was creating the intended sound, and didn't want apps coming to rely on the sound if it turned out to be wrong. If I'm going to add it back, I'll need some convincing that it really is correct or very close.

> BTW, is the reverb echo stage implemented in the current code?

No, for the same reason. I'm doubting the code it had, that it's intended to converge the late reverb taps on the echo point, which messes up the apparent density. I'll need some examples of what it's supposed to sound like and how it influences the late reverb response.

Raulshc:
> Are there any examples of what it sounds like on real hardware, or any other wide-spread reverb implementation? I removed this code because I had serious doubts that it was creating the intended sound, and didn't want apps coming to rely on the sound if it turned out to be wrong. If I'm going to add it back, I'll need some convincing that it really is correct or very close.

Of course, I've attached a zip with a couple examples running in the real hardware [(modulations.zip)](https://github.com/kcat/openal-soft/files/3824691/modulations.zip). The dialtone is taken from [wikimedia commons](https://commons.wikimedia.org/wiki/File:1TR110-1_Kap8.1_Waehlton.ogg), the soft air guitar is obtained from [here](https://samplefocus.com/samples/air-soft-guitar).

The vibrato (phase or pitch modulation) is very common for (electric) guitar musicians. There are a vast range of effect pedals that can be combined. A combination of reverb+phase modulation (discrete or all in one)  is used too.

Googling a lot, you can see descriptions of this:

- [https://guitarlessonssarasota.com/modulated-reverb-part-1/modulated-reverb-part-1/](https://guitarlessonssarasota.com/modulated-reverb-part-1/modulated-reverb-part-1/)
- [https://guitarlessonssarasota.com/modulated-reverb-part-2/modulated-reverb-part-2/](https://guitarlessonssarasota.com/modulated-reverb-part-2/modulated-reverb-part-2/)
- [https://www.guitarlessons.org/gear/pedals/best-reverb-pedals/](https://www.guitarlessons.org/gear/pedals/best-reverb-pedals/)

You can watch several videos:

- [https://www.youtube.com/watch?v=JS2yEznZKDA](https://www.youtube.com/watch?v=JS2yEznZKDA)
- [https://www.youtube.com/watch?v=C9DfYlCBFsk](https://www.youtube.com/watch?v=C9DfYlCBFsk)
- [https://www.youtube.com/watch?v=wzzG03vIDTY](https://www.youtube.com/watch?v=wzzG03vIDTY)

During the development and programming of the current reverb engine you have been able to verify that this "modulation" ( and that "echo" too) have no equivalence in a mathematical description of a reverberation theoretical model. This is because these two stages were added to use them as complements to the reverb effect for guitars and other instruments.

> No, for the same reason. I'm doubting the code it had, that it's intended to converge the late reverb taps on the echo point, which messes up the apparent density. I'll need some examples of what it's supposed to sound like and how it influences the late reverb response.

I will investigate it then. It's a pending task.

kcat:
I just reimplemented this in a different way, which should have better performance but still sound the same. If you notice any problems or discrepancies, please open an issue about it.

Raulshc:
@kcat , that's good. If you need more info of the modulation in the late reverberation, you can see this master thesis: [http://pages.infinit.net/jfrenett/thesis.pdf](http://pages.infinit.net/jfrenett/thesis.pdf)

Thanks

Issue #187, opened 2018-04-16
License question


amerkoleci:
Hi,
And sorry if this is bump question, I would like to know how would openal-soft integrate onto other projects.
I'm working on game engine, would love to add openal-soft under ThirdParty and will also mention the usage into README.md, is that legal?

The engine is MIT and usually build as static library.

Thanks in advance.

kcat:
Hi.

Generally speaking, as long as OpenAL Soft is unmodified and distributed as a shared library, you can use it in any project you wish. Unless your code is also (L)GPL, that's typically the easiest way to deal with LGPL libraries. Otherwise, if OpenAL Soft itself is static-linked into a binary, then any distribution must also offer the source or object files of the binary so a user can relink with another version of OpenAL Soft.

crazyhappygame:
Is there any chance to change licence to MIT? For game development static-linking is very important.

iSLC:
@crazyhappygame I doubt he can. The original licensee would be [Creative Technology](https://en.wikipedia.org/wiki/Creative_Technology). He simply maintains the project. Which is allowed because it satisfies LGPL.

crazyhappygame:
@kcat Did you have a chance to work on "restructuring the library"

kcat:
@crazyhappygame A bit, though this upcoming release will be more focused on the conversion to C++ and making sure it's not causing problems. The restructuring that has been done, along with the switch to C++, should help make the remaining work quicker, though.

halx99:
What about MIT license, recently, I integrate to the game engine fork of cocos2d-x: https://github.com/c4games/engine-x
And I receive message from user of cocos2dx forum: https://discuss.cocos2d-x.org discuss with me about engine audio backend library, I think this lib openal-soft is the best choice for unify audio engine of engine-x for all paltforms, but the LGPL license may limit commercal use for game developer.

McSinyx:
> the LGPL license may limit commercial use for game developer.

Please elaborate.  LGPL only requires source publication of *modification* of, in this case, openal-soft.  Static linking to proprietary software is permitted.

[...]

PR #418, opened 2020-04-20
Update to LGPL v3.0

ChristianFrisson:
Hi @kcat,

Would it be possible to update the license from LGPL v2.0 to v3.0?

Benefits: 
(quoting: https://stackoverflow.com/a/44701 from https://stackoverflow.com/questions/41460/what-are-the-differences-between-gpl-v2-and-gpl-v3-licenses)
- internationalization: they used new terminology, rather than using language tied to US legal concepts
- patents: they specifically address patents (including the Microsoft/Novell issue noted in another answer)
- “Tivo-ization”: they address the restrictions (like Tivo’s) in consumer products that take away, through hardware, the ability to modify the software
- DRM: they address digital rights management (which they call digital restrictions management)
- compatibility: they address compatibility with some other open source licenses
- termination: they address specifically what happens if the license is violated and the cure of violations

Related openal-soft issue: #187 

Kind regards,
Christian

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,437 +1,165 @@
-                  GNU LIBRARY GENERAL PUBLIC LICENSE
-                       Version 2, June 1991
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
- Copyright (C) 1991 Free Software Foundation, Inc.
- 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
-[This is the first released version of the library GPL.  It is
- numbered 2 because it goes with version 2 of the ordinary GPL.]
 
-                            Preamble
+  This version of the GNU Lesser General Public License incorporates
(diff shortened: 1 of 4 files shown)


McSinyx:
I think if @kcat agrees with the change, you'd need to modifies sources' headers too.  [A simple search](https://github.com/kcat/openal-soft/search?q=General+Public+License&type=) may gives you these source files.

kcat:
This could be a problem for apps on Android, iOS, and certain other platforms that restrict user modification of files because of code signing. People do use OpenAL Soft on them, and while I certainly don't like that those systems restrict modifying files, being LGPL v3 would just drive developers away if they can't use OpenAL Soft on them anymore.

Issue #187 is actually about getting a more permissive license, like MIT/Expat, rather than one that restricts use on those kinds of closed systems. Maybe if I get to a point where I can relicense the library as a whole, I can think about possibly relicensing to LGPL v3 with alternative options for platforms that don't allow code with that license. But that would be a whole other discussion.

That said, if you're distributing something and want to include OpenAL Soft under the LGPL v3 for the clarity and such it provides, you can do that since it's given under "version 2 of the License, or (at your option) any later version." I just don't think it would be a good idea for the main release.

ChristianFrisson:
Thanks @McSinyx and @kcat for your recommendations and clarifications. I'll close the pull request.

Issue #311, opened 2019-07-09
Cannot compile on macOS: error: use of undeclared identifier 'aligned_alloc'


hoechenberger:
When trying to compile on macOS 10.14.5 with the latest XCode command line tools installed, I get the following error:
```
$ make
Scanning dependencies of target native-tools
[  1%] Generating native-tools/bin2h, native-tools/bsincgen
-- The C compiler identification is AppleClang 11.0.0.11000020
-- The CXX compiler identification is AppleClang 11.0.0.11000020
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pow in m
-- Looking for pow in m - found
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/hoechenberger/Development/openal-soft/build/native-tools
Scanning dependencies of target bsincgen
[ 25%] Building C object CMakeFiles/bsincgen.dir/bsincgen.c.o
[ 50%] Linking C executable bsincgen
[ 50%] Built target bsincgen
Scanning dependencies of target bin2h
[ 75%] Building C object CMakeFiles/bin2h.dir/bin2h.c.o
[100%] Linking C executable bin2h
[100%] Built target bin2h
[  1%] Built target native-tools
Scanning dependencies of target build_version
[  1%] Built target build_version
[  2%] Generating bsinc_inc.h
[  4%] Generating default-44100.mhr.h
[  5%] Generating default-48000.mhr.h
Scanning dependencies of target OpenAL
[  7%] Building CXX object CMakeFiles/OpenAL.dir/common/alcomplex.cpp.o
[  8%] Building CXX object CMakeFiles/OpenAL.dir/common/alexcpt.cpp.o
[ 10%] Building CXX object CMakeFiles/OpenAL.dir/common/almalloc.cpp.o
[...]

PR #426, opened 2020-05-15
Add HAVE_ALIGNED_ALLOC check back for some broken c-runtime

taigacon:
Some broken c-runtime lacks the aligned_alloc function even if the compiler supports C11/C++17 (by passing -std=c++17 to compiler).
For example, if someone builds the project and targets to Android below API 28, or to iOS below iOS 13, or to macOS below macOS 10.11, the compile or link will fail.


CMakeLists.txt
```diff
@@ -449,6 +449,7 @@ IF(HAVE_INTRIN_H)
         }" HAVE_BITSCANFORWARD_INTRINSIC)
 ENDIF()
 
+CHECK_SYMBOL_EXISTS(aligned_alloc   stdlib.h HAVE_ALIGNED_ALLOC)
```
kcat, line 452: This unfortunately only tests under C, and some systems may expose `aligned_alloc` for C11, but not for C++ prior to C++17. And even if it is available in C++, it's technically `std::aligned_alloc`, just `aligned_alloc` isn't guaranteed to work (in practice it probably will, but I don't think it has to).

taigacon: Yes, it's right. However, almalloc.cpp uses `aligned_alloc` in C, not `std::aligned_alloc` in CPP.
So I think that if we make it more strict, we should check c++17 flag and check the existence of `std::aligned_alloc` function, then use `std::aligned_alloc` instead, or checking then using `aligned_alloc` in `stdlib.h` may be enough.

kcat:
> Yes, it's right. However, almalloc.cpp uses `aligned_alloc` in C, not `std::aligned_alloc` in CPP.

I'll need to fix that, since `almalloc.cpp` is compiled as C++ instead of C.

> So I think that if we make it more strict, we should check c++17 flag and check the existence of `std::aligned_alloc` function, then use `std::aligned_alloc` instead, or checking then using `aligned_alloc` in `stdlib.h` may be enough.

Checking and using `aligned_alloc` in `stdlib.h` unfortunately doesn't work on all systems (see issue #311). Checking for `__cplusplus >= 201703L` should be enough to test if C++17 is being used, which has `std::aligned_alloc`. I'm curious if changing `aligned_alloc` to `std::aligned_alloc` would be enough to make Android/iOS/macOS happy, if they're being more strict with not exposing C functions in C++.

taigacon:
> Checking and using `aligned_alloc` in `stdlib.h` unfortunately doesn't work on all systems (see issue #311). Checking for `__cplusplus >= 201703L` should be enough to test if C++17 is being used, which has `std::aligned_alloc`. I'm curious if changing `aligned_alloc` to `std::aligned_alloc` would be enough to make Android/iOS/macOS happy, if they're being more strict with not exposing C functions in C++.
We may need to check and use c++17 first, then check the existence of `std::aligned_alloc`. `std::aligned_alloc` is not valid on some platforms even if the compiler enables c++17 mode, such as MSVC.

> https://devblogs.microsoft.com/cppblog/announcing-msvc-conforms-to-the-c-standard/
> [C11]: First available in VS 2015, the Universal CRT implemented the parts of the C11 Standard Library that are required by C++17, with minor exceptions. Those exceptions (which are tracked by bugs) are: missing C99 strftime() E/O alternative conversion specifiers, missing C11 fopen() exclusive mode, and missing C11 aligned_alloc(). The strftime() and fopen() functionality will be implemented in the future. aligned_alloc() will probably never be implemented, as C11 specified it in a way that’s incompatible with our implementation (namely, that free() must be able to handle highly aligned allocations).

kcat:
I suppose an alternative to using `std::aligned_alloc` is to use the aligned new, i.e.
`::new(std::align_val_t(alignment)) std::byte[size];`
Though if the system supports that (Android NDK r16 should, from what I'm reading; I don't know which versions of iOS and macOS claim C++17 compatibility), I don't know why it wouldn't also support `std::aligned_alloc`, MSVC not withstanding.

taigacon:
According to [https://reviews.llvm.org/D45015](https://reviews.llvm.org/D45015), at lease clang will use `aligned_alloc` to implement `::new(std::align_val_t(alignment))` if the target specifies `-faligned-alloc`, and will use a fallback version (malloc + offset) if no `aligned_alloc` is valid ( or target specifies `-fno-aligned-alloc` ).
Here is the `aligned_alloc` definition in iOS SDK:
`#if (__DARWIN_C_LEVEL >= __DARWIN_C_FULL) && \
        ((defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L) || \
         (defined(__cplusplus) && __cplusplus >= 201703L))
void    *aligned_alloc(size_t __alignment, size_t __size) __result_use_check __alloc_size(2) __OSX_AVAILABLE(10.15) __IOS_AVAILABLE(13.0) __TVOS_AVAILABLE(13.0) __WATCHOS_AVAILABLE(6.0);
#endif`

`aligned_alloc` is valid only after iOS 13.0 or tvOS 13.0 or watchOS 6.0 or macOS 10.15.
And `aligned_alloc` is valid after Android API-28.
Old devices will never support `aligned_alloc` due to their outdated c-runtime in their systems.

`#if _LIBCPP_STD_VER > 14 && defined(_LIBCPP_HAS_C11_FEATURES)
using ::aligned_alloc;
#endif`

`std::aligned_alloc` is just the alias of `aligned_alloc` in libc++ with some conditions.

So I think we should check and enable c++17 mode first in CMake, and then check `std::aligned_alloc` method.

MSVC has some problems left over by history:
[https://developercommunity.visualstudio.com/content/problem/468021/c17-stdaligned-alloc%E7%BC%BA%E5%A4%B1.html](https://developercommunity.visualstudio.com/content/problem/468021/c17-stdaligned-alloc%E7%BC%BA%E5%A4%B1.html)

> The reason for this difference is due to the implementation of malloc and free in the Universal C Runtime. The Universal C Runtime uses the Win32 Heap APIs directly to implement malloc() and free(), however the Win32 Heap does not support allocations with specific alignment. To provide an aligned allocation via _aligned_malloc, the Universal C Runtime over-allocates and then returns a pointer within the space that was allocated that is aligned, and keeps extra bookkeeping about this allocation. The free() function would free the pointer it was given, but for an aligned allocation this means we have to find the larger block that was allocated and free that one instead along with the extra bookkeeping data, creating the need for a different function.

kcat:
> `aligned_alloc` is valid only after iOS 13.0 or tvOS 13.0 or watchOS 6.0 or macOS 10.15.
> And `aligned_alloc` is valid after Android API-28 (Android 9.0).
> Old devices will never support `aligned_alloc` due to their outdated c-runtime in their systems.

Right, but I mean, how those versions fit in with C++17 compliance in general. If those versions were among the first to have official support for `-std=c++17`, then it makes sense that certain C++17 features may be missing before then because C++17 support was incomplete before then (and consequently, code aiming to be C++17-compliant shouldn't hold to compatibility with pre-C++17-compliant compilers/systems; that defeats the purpose of using the C++17 standard to not be able to use the C++17 standard).

However, if there was a significant delay between compilers becoming C++17-compliant and the relevant OS target becoming C++17-compliant, it may be worth having a temporary workaround for those OS targets until support becomes more widespread.

>> To provide an aligned allocation via _aligned_malloc, the Universal C Runtime over-allocates and then returns a pointer within the space that was allocated that is aligned, and keeps extra bookkeeping about this allocation.

Wow, and I thought my opinion of Windows/MSVC was already low. Its allocator doesn't even have real alignment capabilities, it just does a malloc+offset (with requisite bookkeeping) no better than user code does. Might not even be worth using it then, since OpenAL Soft already has a malloc+offset fallback. Glad I didn't rely on it for 4KB page-aligned blocks, thinking it would efficiently allocate them.

taigacon:
>Right, but I mean, how those versions fit in with C++17 compliance in general.

According to [https://libcxx.llvm.org/docs/DesignDocs/AvailabilityMarkup.html](https://libcxx.llvm.org/docs/DesignDocs/AvailabilityMarkup.html), clang and libc++ is more stable than we imagine. All APIs have their minimum supported system version on the Apple family. All functionality can be checked and tested in the build time.
Android NDK uses clang and libc++ default too. But libc++ is not embedded in the Android system, the application will carry a binary copy of libc++ in the release package (apks), so nothing needs to be worried about.

>However, if there was a significant delay between compilers becoming C++17-compliant and the relevant OS target becoming C++17-compliant, it may be worth having a temporary workaround for those OS targets until support becomes more widespread.

Yes, there is a significant delay nowadays. The best proof is Clang now use C++14 default.

>Or I could just not bother with std::aligned_alloc at all, even if C++17 is detected. 

I think `aligned_alloc` is meaningful for Linux or specific system, when building from source only. If we distribute the binary form of our library, we may not use C++17/C11 library function until they become more widespread.
`posix_memalign` is enough to use in almost all cases.
We can simply check the existence of `std::aligned_alloc` in configure time. If the compiler defaults to using C++17, and then compiling and linking `std::aligned_alloc` succeed,  I think it is safe to use. 

> Wow, and I thought my opinion of Windows/MSVC was already low.

It works at least:)

taigacon:
By the way, I think the reason for issue #311 is that Clang defaults to use -std=gnu11 for C and use -std=c++14 for C++. `aligned_alloc` actually exists in C11 and actually not exists in C++14.
So we just change CHECK_C_SOURCE_COMPILES to CHECK_CXX_SOURCE_COMPILES, everything will be fine.

kcat:
Commit 463591663c2421b3436edc446d173380d6a6e106 re-adds the check using `check_cxx_source_compiles` instead of `check_symbol_exists`.

taigacon:
Great, and this PR can be closed.

PR #435, opened 2020-06-07
sndio: Support more than 2 channels

DanielGibson:
So far the sndio backend only supports Mono and Stereo, but sndio
supports up to 16 channels (64 in their current git code).
This adds support for all the openal-soft formats (DevFmt*).
I tested this on Linux with 5.1 surround speakers.

As a sidenode, it might make sense to mention sndio *somewhere* in the docs or alsoftrc.sample to people know it's supported and know how the driver is called for `drivers = ...` and `ALSOFT_DRIVERS` (for the other backends it can be looked up in their alsoftrc config section, but sndio doesn't have one as it doesn't seem to have configurable options).

### DIFF
--- a/alc/backends/sndio.cpp
+++ b/alc/backends/sndio.cpp
@@ -127,7 +127,21 @@ bool SndioPlayback::reset()
     sio_initpar(&par);
 
     par.rate = mDevice->Frequency;
-    par.pchan = ((mDevice->FmtChans != DevFmtMono) ? 2 : 1);
+    switch(mDevice->FmtChans)
+    {
+        case DevFmtMono   : par.pchan = 1; break;
+        case DevFmtQuad   : par.pchan = 4; break;
+        case DevFmtX51Rear: // fall-through - "Similar to 5.1, except using rear channels instead of sides"
+        case DevFmtX51    : par.pchan = 6; break;
+        case DevFmtX61    : par.pchan = 7; break;
+        case DevFmtX71    : par.pchan = 8; break;
+
+        case DevFmtStereo : // fall-through
+        default: // fall back to stereo for all unknown formats and Ambi3D
+            mDevice->FmtChans = DevFmtStereo;
+            par.pchan = 2;
+            break;
+    }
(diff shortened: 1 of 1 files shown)


kcat:
What is the sndio API behavior in this regard? If you try to request 6 channels, for example, does it guarantee to give back 6 channels, or might it be automatically changed to anything else that may be available, like 4, 5, 7, 8, etc? And does sndio guarantee specific meaning for particular channel counts (e.g. is `4` guaranteed to be quadraphonic and `6` be 5.1 surround)? Or is there perhaps an alternate API that can handle channel layouts instead of channel counts?

DanielGibson:
As far as I understand the documentation (http://man.openbsd.org/sio_open) it will return something that's available and the ALSA-backend uses `snd_pcm_hw_params_set_channels_near()` to get the real channel count.

I don't think there's any information about the speaker layout available, so I think matching the number of channels is the best we can get. I think this isn't that bad, someone using sndio should be able to configure the `channels` option in the alsoft config in a way that matches their setup (and I think even for ALSA one has to set that option to get something else than stereo?).

kcat:
> I think this isn't that bad, someone using sndio should be able to configure the `channels` option in the alsoft config in a way that matches their setup (and if they don't configure it they just get stereo which should almost always work - I think even for ALSA one has to set that option to get something different from stereo?).

As far as setting channels goes, yeah. The lib defaults to stereo, and if the user configures it to surround51 or something, we can assume the channel count means the desired layout. However, it's good to also have contingencies in place in case something unexpected comes back. OpenAL Soft tries pretty hard to ensure something usable gets configured for playback, so it's also not good to just fail if it can otherwise be helped.

Luckily, the first two channels of any multi-channel layout being front-left and front-right is a fairly safe assumption, and since 1.20.1, it's been possible to use a consecutive subset of output channels. So there are ways to deal with it, I was just curious what to expect and what options there are.

DanielGibson:
Thank you! :-)

PR #446, opened 2020-06-26
dirty way to solve mingw64 compile error

LemonHX:


kcat:
The issue is already fixed with commit 31791c9997d06a3a315d2b936515f643430cd5e9.

Issue #406, opened 2020-03-31
Is iOS supported? If so, how do you compile it?


jay072:
I've seen a few issues on this repository regarding iOS support, but it's not clear if it's officially supported, and I haven't found any instructions on how to set it up. I cannot use the native OpenAL framework anymore, since Apple has deprecated this in the latest iOS and is likely to be removed soon.

Is it possible to build for iOS, if so: how?

I've successfully used the following toolchain to setup the Xcode project
https://github.com/cristeab/ios-cmake/blob/master/toolchain/iOS.cmake

However, when building I get issues about the "native-tools" target.
```target specifies product type 'com.apple.product-type.tool', but there's no such product type for the 'iphoneos' platform```

```
cd /.../openal-soft/ios/native-tools && /Applications/CMake.app/Contents/bin/cmake -G Xcode /.../openal-soft/native-tools
-- The C compiler identification is AppleClang 11.0.3.11030032
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -- broken
CMake Error at /Applications/CMake.app/Contents/share/cmake-3.16/Modules/CMakeTestCCompiler.cmake:60 (message):
  The C compiler

    "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /.../openal-soft/ios/native-tools/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -project CMAKE_TRY_COMPILE.xcodeproj build -target cmTC_7b571 -configuration Debug -hideShellScriptEnvironment && Command line invocation:
        /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -project CMAKE_TRY_COMPILE.xcodeproj build -target cmTC_7b571 -configuration Debug -hideShellScriptEnvironment
    
    User defaults from command line:
        HideShellScriptEnvironment = YES
    
    Build settings from command line:
        TOOLCHAINS = com.apple.dt.toolchain.XcodeDefault
    
    Prepare build
    note: Using legacy build system
[...]

PR #452, opened 2020-07-21
Add ios build for travis

halx99:


kcat:
For some reason it seems to not be recognizing NEON extensions, but does get SSE and `cpuid`. Doesn't iphone use ARM? Why would it be recognizing x86 things?

halx99:
because previous commit, I use iphonesimulator as sysroot, and after I change to ```iphoneos```, just failed as I said: https://github.com/kcat/openal-soft/issues/406

halx99:
I have test this https://github.com/leetal/ios-cmake toolchan, it works.
But I meet crash issue when I try to integrate latest cocos2d-x v4.
and I realized the openal latest reqiured c++14, but the cocos2dx v4 default c++ standard is 11
may be the different compiler cause the crash, I will continue test later

halx99:
I have tested, openal-soft works fine on both ios simulator and devices, so I think this PR could be merge

CMakeLists.txt
```diff
@@ -134,6 +134,7 @@ check_type_size("long" SIZEOF_LONG)
 # Require C++14
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
+set(CMAKE_CXX_EXTENSIONS FALSE)
```
kcat, line 137: I don't think I like this. It inhibits optional but beneficial features that could be useful for the generated code.

halx99: agree with you, removed

PR #453, opened 2020-07-21
GetProcBinary implementation for Haiku.

devnexen:


alc/helpers.cpp
```diff
@@ -255,6 +258,13 @@ const PathNamePair &GetProcBinary()
         else
             pathname.insert(pathname.end(), procpath, procpath+strlen(procpath));
     }
+#endif
+#ifdef __HAIKU__
+    char procpath[PATH_MAX];
+    if(find_path(B_APP_IMAGE_SYMBOL, B_FIND_PATH_IMAGE_PATH, NULL, procpath, sizeof(procpath)) == B_OK)
```
kcat, line 264: Does this get the full path and executable name (e.g. `/usr/bin/myapp`), or just the path the executable is in (e.g. `/usr/bin`)? The function and enum names suggest the latter, and [the docs I found](https://www.haiku-os.org/docs/api/FindDirectory_8h.html) don't really clarify.

devnexen: Nope it really returns the full path exe binary included. Agreed most of Haiku man/docs do not always reflect the reality or lack clarity, had to figure out most of things during porting softwares :-)

PR #458, opened 2020-07-27
JACK should not automatically connect to outputs

Geo25rey:
The whole point of JACK is to customize your audio inputs and outputs, so OpenAL shouldn't change any JACK connections.

### DIFF
--- a/alc/backends/jack.cpp
+++ b/alc/backends/jack.cpp
@@ -385,29 +385,6 @@ void JackPlayback::start()
     if(jack_activate(mClient))
         throw al::backend_exception{ALC_INVALID_DEVICE, "Failed to activate client"};
 
-    const char **ports{jack_get_ports(mClient, nullptr, nullptr,
-        JackPortIsPhysical|JackPortIsInput)};
-    if(ports == nullptr)
-    {
-        jack_deactivate(mClient);
-        throw al::backend_exception{ALC_INVALID_DEVICE, "No physical playback ports found"};
-    }
-    std::mismatch(std::begin(mPort), std::end(mPort), ports,
-        [this](const jack_port_t *port, const char *pname) -> bool
-        {
-            if(!port) return false;
-            if(!pname)
-            {
-                ERR("No physical playback port for \"%s\"\n", jack_port_name(port));
-                return false;
-            }
-            if(jack_connect(mClient, jack_port_name(port), pname))
-                ERR("Failed to connect output port \"%s\" to \"%s\"\n", jack_port_name(port),
-                    pname);
-            return true;
-        });
-    jack_free(ports);
-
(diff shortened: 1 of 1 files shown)


kcat:
Won't this mean someone using JACK for output will always get silence until they manually connect the channels to the outputs?

Geo25rey:
@kcat that's kinda the point of JACK

leafi:
I don't think most users of OpenAL Soft even know what JACK is. If it happens to be chosen automatically as the output API for an OpenAL-using video game - say, because Pipewire eventually replaces PulseAudio, and Pipewire has both ALSA, PulseAudio and JACK library shims - the game or application having to add JACK-specific code to actually connect generated sound to the outputs kinda defeats the point of using an audio abstraction layer like OpenAL that hides all the dirty details.

I've only used a few JACK apps, but Renoise, Audacity, and REAPER, all DAWs, seem perfectly happy to connect themselves to speaker outputs by default. Or maybe something else auto-connects them: I'm not sure which layer this happens at.

Geo25rey:
There is software like qjackctl which can automatically attach inputs and outputs to what you want with a patchbay. 

In my setup, I use an extra set of JACK ports to act as a master software volume. This JACK connection uses software input and outputs, but, as of currently, OpenAL connects automatically to physical outputs and thus, kills my ears until I manually disconnect the connection.

okuoku:
How about having `autoconnect` option instead of removing the feature. For example [gstreamer](https://gstreamer.freedesktop.org/documentation/jack/jackaudiosink.html?gi-language=c#jackaudiosink) (default ON), [fluidsynth](http://www.fluidsynth.org/api/fluidsettings.xml#audio.jack.autoconnect) (default OFF).

Geo25rey:
@okuoku I'm not sure how that would be implemented, since, to my knowledge, OpenAL doesn't provide any interface for directly changing options of the sound backends. Perhaps you mean a compile time option?

okuoku:
@Geo25rey : openal-soft has `alsoft.conf` or [environment variables](https://github.com/kcat/openal-soft/blob/7cac5304c8e4a70ac0188693dbcb6eaf46b7601f/docs/env-vars.txt) to configure its behavior. It also has Qt app `alsoft-conf` as user interface.

kcat:
> How about having `autoconnect` option instead of removing the feature.

This would make the most sense to me. Commit f138369808086cbf2bdf1d579e6f21553c657dc5 adds the `connect-ports` option under the `[jack]` header (defaults to true), and commit 7ddd49e16a3ac854eaa88f2e3367575bf6d5b7ec adds it to `alsoft-config`.

Geo25rey:
Thanks @kcat

PR #469, opened 2020-09-06
Added the possibility to change output device with JACK

Vahlblue:
This modification change the way the JACK backend choose wich ports to connect the same way VLC do to allow the choice of the output device :
If no device is given it choose to start on "system", the default JACK ouput.
If a device is specified, it use jack_get_ports to enumerate ports.
For example if device="ardour:AL/audio_in", it will connect to : 
ardour:AL/audio_in 1
ardour:AL/audio_in 2

For JackBackendFactory::probe, I've still have an hesitaion :
JACK can enumerate all ports, however it can't list devices.
Shoud JackBackendFactory::probe give :
- only "system" ?
- all ports (currently what my modification do) ?
- JACK client names ? (ex : "Ardour", "system") But some clients (like Ardour) have too many ports, it will not connect to the correct ports.
- or we try recover port groups (ex : "ardour:AL/audio_in") But for this solution, we need to list how channels can be named by clients (Ardour, alsa_out and JACK use 1/2 ; JACK mixer uses L/R ; the JACK sink Pulseaudio module uses left/right...).

### DIFF
--- a/alc/backends/jack.cpp
+++ b/alc/backends/jack.cpp
@@ -483,14 +483,14 @@ void JackPlayback::start()
         if(ports == nullptr)
         {
             jack_deactivate(mClient);
-            throw al::backend_exception{ALC_INVALID_DEVICE, "No physical playback ports found"};
+            throw al::backend_exception{ALC_INVALID_DEVICE, "No playback ports found"};
         }
         auto connect_port = [this](const jack_port_t *port, const char *pname) -> bool
         {
             if(!port) return false;
             if(!pname)
             {
-                ERR("No physical playback port for \"%s\"\n", jack_port_name(port));
+                ERR("No playback port for \"%s\"\n", jack_port_name(port));
                 return false;
             }
             if(jack_connect(mClient, jack_port_name(port), pname))
--- a/alc/backends/jack.cpp
+++ b/alc/backends/jack.cpp
@@ -167,6 +167,32 @@ void EnumerateDevices(al::vector<DeviceEntry> &list)
 
     list.emplace_back(DeviceEntry{jackDevice, ""});
 
+    jack_client_t *client{nullptr};
+    jack_status_t status;
+    client = jack_client_open("alsoft", JackNoStartServer, &status, nullptr);
+    if(client == nullptr)
+        throw al::backend_exception{ALC_INVALID_VALUE, "Failed to open client connection: 0x%02x",
+                status};
+    const char **ports{jack_get_ports(client, nullptr, nullptr, JackPortIsPhysical|JackPortIsInput)};
+    for (unsigned i = 0; ports && ports[i] ; i++)
+    {
+        std::string pattern{ports[i]};
+        size_t colon = pattern.find(':');
+        if (colon != std::string::npos && colon != 0)
+        {
+            pattern.resize(colon + 1);
+            std::string name = pattern;
+            name.pop_back();
+            DeviceEntry device = {name, pattern};
+            auto same_pattern = [&pattern](const DeviceEntry &entry) -> bool
+            { return entry.mPattern == pattern; };
+            if (std::find_if(list.begin(), list.end(), same_pattern) == list.end())
+                list.emplace_back(device);
+        }
+    }
+    jack_free(ports);
+    jack_client_close (client);
+
     std::string customList{ConfigValueStr(nullptr, "jack", "custom-devices").value_or("")};
     size_t strpos{0};
     while(strpos < customList.size())
--- a/alc/backends/jack.cpp
+++ b/alc/backends/jack.cpp
@@ -152,8 +152,91 @@ bool jack_load()
 }
 
 
+static void update_list_devices(std::vector<std::string> &device_list_names, std::vector<std::string> &device_list)
+{
+    device_list_names.clear();
+    device_list.clear();
+    jack_client_t *client{nullptr};
+    jack_status_t status;
+    client = jack_client_open("alsoft", JackNoStartServer, &status, nullptr);
+    if(client == nullptr)
+        throw al::backend_exception{ALC_INVALID_VALUE, "Failed to open client connection: 0x%02x",
+                status};
+    const char **ports{jack_get_ports(client, nullptr, nullptr, JackPortIsPhysical|JackPortIsInput)};
+    for (unsigned i = 0; ports && ports[i] ; i++)
+    {
+        const char* colon = strchr(ports[i], ':');
+        size_t len;
+        if (colon != nullptr)
+        {
+            len = static_cast<size_t>  (colon - ports[i]) + 1;
+        }
+        else
+        {
+            len = strlen(ports[i]);
+        }
+        bool isInList = false;
+        unsigned ii = 0;
+        while (!isInList && ii != device_list_names.size())
+        {
+            isInList = (device_list_names[ii].compare(0, std::string::npos, ports[i], len - 1) == 0);
+            ii++;
+        }
+        if (!isInList)
+        {
+            device_list_names.push_back(std::string(ports[i], len - 1));
+            device_list.push_back(std::string(ports[i], len));
+        }
+    }
+    jack_free(ports);
+    jack_client_close (client);
+
+    char *customDevices = strdup(GetConfigValue(nullptr, "jack", "custom-devices", ""));
+    if (strlen(customDevices) != 0)
+    {
+        char* end = strchr(customDevices, '\0');
+        char *ptr = customDevices;
+        while (ptr != (end + 1))
+        {
+            char *ptrnext = strchr(ptr, ';');
+            if (ptrnext == nullptr)
+            {
+                ptrnext = end;
+            }
+            size_t len = static_cast<size_t>  (ptrnext - ptr);
+            if (len > 0)
+            {
+                std::string device, device_name;
+                const char* deviceptr = strchr(ptr, '=');
+                size_t lenname = static_cast<size_t> (deviceptr - ptr);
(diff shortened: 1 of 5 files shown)


alc/backends/jack.cpp
```diff
@@ -503,11 +500,29 @@ bool JackBackendFactory::querySupport(BackendType type)
...
+        if(client == nullptr)
+            throw al::backend_exception{ALC_INVALID_VALUE, "Failed to open client connection: 0x%02x",
+                    status};
+        const char **ports{jack_get_ports(client, nullptr, nullptr, JackPortIsInput)};
+        for (int i = 0; ports && ports[i] ; i++) {
+            /* +1 to also append the null char (to ensure a null-separated list and
+            * double-null terminated list).
+            */
+            outnames.append(ports[i], strlen(ports[i]) + 1);
+        }
```
kcat, line 517: Won't this report each port/channel as its own device? On my system, this gives `system:playback_1`, `system:playback_2`, etc, for each playback channel.

Vahlblue: Yes that's intended (for now).
It's patially flawed because I don't see a perfect solution to recover easyly the real outputs.
That's one of the 4 options I see to implement JackBackendFactory::probe (I implemented for now this one, because it's simple to implement and at least gives more possibility to to the user).
However, I don't know wich one would be the best.
- Return just "system" : this would break nothing, but still give the user the possibility to change manualy the output (ex : by changing the config file in OpenMW)
- Return all ports (as it is) : certainly not the best, simple to implement but will certainly break things
- Return JACK client names : give the user to choose beetween client freely (physicals output and simple clients will work). But some clients, like Ardour, will not work. But it will give the possibility to the user to choose "system" if nothing works.
- Ruturn the recovered real ouputs names : with the list of all ports, by cutting the end of the names by reconnizing some patterns (L/R, right/left, 1/2 or alternativly cut the last word after a '_' or a '-'). It will work on most clients and with jack's physical outputs, but may not work with some clients. I think this one is the best but it may confuses the user if he want to understand why their client doesn't work.

Which one do you think would be the best ?

kcat: > It's patially flawed because I don't see a perfect solution to recover easyly the real outputs.

The main problem I see is that JACK doesn't really distinguish outputs. It just presents ports that each represent a playback channel to whatever devices are on the system (or act as a source to something else). There's no introspection to discover JACK devices (real or virtual), with the ports that belong to those specific devices. Consequently, the user is responsible for setting anything up.

One thought I had was looking through the ports and taking everything up to `:` as the device name (e.g. `system` for one device), although I don't know if that's a guaranteed pattern, or if different devices even have a different name before the colon. Cutting off the last word after ` `, `_`, or `-` and using the remaining as a name/pattern also feels sketchy and error-prone.

I think the more prudent option would be to allow the user to configure their own set of device names and port patterns in the config file. Something like:
```
[jack]
custom-devices = My Device=some_ports*;Another Device=other_ports*
```
Then OpenAL Soft would enumerate `JACK Default`, `My Device`, and `Another Device`, which would use the port name filters ``, `some_ports*`, and `other_ports*` respectively.

Vahlblue: >  although I don't know if that's a guaranteed pattern

Jack seems to delimit the device name and the "short" port name by the first `:`. So no problem with that.
[https://github.com/jackaudio/jack2/blob/98c882c5a545c3dd819655c9869f6928274df8a3/common/JackPort.cpp#L224](url)

>  I think the more prudent option would be to allow the user to configure their own set of device names and port patterns in the config file.

In this case, I think `<blank>` should list device names (`system`, `alsa_out`, ...) corresponding to physical outputs. So  every users may connect simply to the output of their choice (not all devices, because it may confuse unexperienced users if it doesn't work). Then if an advanced user want to use a non physical ports, they may use custom patterns like you said.

Vahlblue:
Now how it behave :
By default `JackBackendFactory::probe` gives the device name list corresponding to physical ports (`system`, `alsa_out`, ...).
If `custom-devices is defined`, each device given is added to the list. So when one is set as the output, `jack_get_ports` will gives all the ports corresponding to the device pattern chosen.

alc/backends/jack.cpp
```diff
@@ -508,15 +508,32 @@ std::string JackBackendFactory::probe(BackendType type)
         if(client == nullptr)
             throw al::backend_exception{ALC_INVALID_VALUE, "Failed to open client connection: 0x%02x",
                     status};
-        const char **ports{jack_get_ports(client, nullptr, nullptr, JackPortIsInput)};
+        const char **ports{jack_get_ports(client, nullptr, nullptr, JackPortIsPhysical|JackPortIsInput)};
         for (int i = 0; ports && ports[i] ; i++) {
+            const char* colon = strchr(ports[i], ':');
+            int len = static_cast<int>  (colon - ports[i]);
+            if (outnames.find(ports[i], 0, len) == std::string::npos) {
```
kcat, line 515: This will break if there isn't a `:` in the port name (not likely to happen, but best to be defensive when dealing with external data). This will also be a problem if something like `alsa_out2` gets added before `alsa_out` (the latter will match against the former and not get added).

alc/backends/jack.cpp
```diff
@@ -508,15 +508,32 @@ std::string JackBackendFactory::probe(BackendType type)
...
+        jack_client_close (client);
+
+        char *customDevices = strdup(GetConfigValue(nullptr, "jack", "custom-devices", ""));
+        size_t len = strlen(customDevices);
+        if (len != 0) {
+            char *ptr = strrchr(customDevices, ';');
+            while (ptr) {
+                *ptr = '\0';
+                ptr = strrchr(ptr, ';');
+            }
```
kcat, line 530: This would need to separate the display name from the port name pattern. It will also need to maintain a list of custom names to know what the desired pattern is when such a name is given for the backend `open` call. And it probably should ensure there are some ports matching the pattern before adding it to the enumerated list.

alc/backends/jack.cpp
Vahlblue: >  This will break if there isn't a `:`

I just added a check if there isn't one : the complete name will be added

 >  alsa_out2 gets added before alsa_out

For this, including the final `:` by adding one to the length should be enough to force the device name. And with the list of custom names, we could remove the `:` afterward for more clarity.

alc/backends/jack.cpp
Vahlblue: >  And it probably should ensure there are some ports matching the pattern before adding it to the enumerated list.

I think it would be better to let everything in the device list. And in the case of no ports corresponding, openal should just be let unconnected. Because if the user want to start a client which is in the `custom-devices` list after the `probe` call, it will be easier for the user to rechange the device choice than trying to figure out how to make the application retrigger `probe` to update the list (reopening the menu ? restarting the whole application ?). I don't think every application using openal would thinks that a device may appear during the run of the application, and make it difficult to update the list, as a device is unlikely to appear for Alsa/OSS/Pulseaudio/... but not for Jack.

Vahlblue:
Now these patterns are accepted :
 - `name=device`
 - `device`

And these are ignored :
 - `=device`
 - `name=`
 - `=`
- `<blank>`

For know, the existence of matching ports isn't checked.

kcat:
I've implemented custom device enumeration as `Name=pattern[;Name=pattern;...]`. These are the patters as given to `jack_get_ports` so should contain a regular expression to get a list (e.g. `ALSA Out Ports=alsa_out:*`  to have an `ALSA Out Ports` device enumerated, that connects to `alsa_out:...` ports if JACK has any).

I'm unsure how I want to deal with automatic enumeration since JACK doesn't have any "pretty" names. It'd just be `system`, `alsa_out`, etc, which isn't very clear. Custom devices allow the user to specify what they want and give it a meaningful name, so I may just leave it with that unless there's a better option.

I'm also wondering if there should be a way for a user to specify a specific port list that a device should try connecting to, rather than using a pattern with the ports in whatever order JACK gives them in.

Vahlblue:
>  specify a specific port list

It will might be useful when there is over 2 channels, as they may not appear in the good order.

> "pretty" names

I just checked the API. And there is maybe something to dig with JACK_METADATA_HARDWARE and JACK_METADATA_PRETTY_NAME metadata proprieties.

Vahlblue:
I just added the automatic enumeration of physical device like this :
for `client-name:*something*` it adds `client-name` as the displayed name and `client-name:` as the pattern.

Vahlblue:
Here is a little summary of this PR, because it was made a long time ago :
- Added automatic discovery of clients
- Allowed the ports to be non-physical (and change the error messages associated with)
- Added the `custom-devices` to  alsoftrc.sample and alsoft-config

Is there anything to change ?

kcat:
I don't think automatic discovery is a good idea, without some way to give good/informative names.

I'll need to look into what the consequences are with allowing non-physical ports, whether it's normal to recognize them by default or if it should only recognize them with custom devices.

I forgot to add `custom-devices` to `alsoftrc.sample`. A UI will need to be designed for `alsoft-config`, which is more than just a text input line that's saved as-is to the config setting. ALSA needs something similar too, actually.

kcat:
Enumeration has been added with commit 698799da0edc40afd094b6f0a7530a1df7837d35. If other apps expect port names to be formatted as `device:port`, I guess that's good enough. PipeWire also seems to give "pretty" device/app names through the JACK API, so at least with that, it won't look ugly. There is no special handling for `system:`, and it doesn't require ports to be physical to consider them. The only thing that's really missing is a config option to indicate a preferred default output, instead of assuming the first is.

PR #477, opened 2020-09-28
Fix syntax error in IRC_1005.def

MatthewScholefield:
When I was trying to convert [IRCAM HRTF data](http://recherche.ircam.fr/equipes/salles/listen/) with openal's `makemhr` utility, it wouldn't work with the default `IRC_1005.def` because the syntax for the definitions must have changed. With the existing `left` or `right` suffix, the parser assumed the file had ended because the next symbol was not `[` which caused it to check if all directions were defined causing the following error:

```
Error (IRC_1005.def:42:83): Missing source reference [ 0, 3, 1 ].
```

By removing the suffix it parses the definition correctly and runs properly. My assumption is that this syntax was removed since it seems to be redundant with the inclusion of the `(0)` or `(1)` in `wave (0) : "..."`.

### DIFF
--- a/utils/IRC_1005.def
+++ b/utils/IRC_1005.def
@@ -39,387 +39,387 @@ azimuths = 1, 6, 12, 24, 24, 24, 24, 24, 24, 24, 12, 6, 1
 
 # Replace all occurrences of IRC_#### for the desired subject (1005 was used
 # in this demonstration).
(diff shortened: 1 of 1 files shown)


MatthewScholefield:
Whoops, I was mistakenly passing in the `-m` flag. Better error messages would be a nice feature though. I might follow up with a PR with documentation of some sort if that would be welcomed.

PR #502, opened 2020-12-04
Add a gain argument to altonegen

eliotmiranda:
Add a gain command-line argument to altonegen.c.
Fix a spelling error in an error message in opensl.cpp
Fix compilation of some of the examples by guarding the include of win_main_utf8.h

### DIFF
--- a/examples/altonegen.c
+++ b/examples/altonegen.c
@@ -44,7 +44,9 @@
 
 #include "common/alhelpers.h"
 
-#include "win_main_utf8.h"
+#ifdef _WIN32
+# include "win_main_utf8.h"
+#endif
 
 #ifndef M_PI
 #define M_PI    (3.14159265358979323846)
@@ -93,7 +95,7 @@ static void ApplySin(ALfloat *data, ALdouble g, ALuint srate, ALuint freq)
 /* Generates waveforms using additive synthesis. Each waveform is constructed
  * by summing one or more sine waves, up to (and excluding) nyquist.
  */
-static ALuint CreateWave(enum WaveType type, ALuint freq, ALuint srate)
+static ALuint CreateWave(enum WaveType type, ALuint freq, ALuint srate, ALfloat gain)
 {
     ALuint seed = 22222;
     ALuint data_size;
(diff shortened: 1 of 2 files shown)


examples/altonegen.c
```diff
@@ -241,6 +251,16 @@ int main(int argc, char *argv[])
...
             }
         }
+        else if(i+1 < argc && (strcmp(argv[i], "--gain") == 0 || strcmp(argv[i], "-g") == 0))
+        {
+            i++;
+            gain = atof(argv[i]);
+            if(gain < 0.0 || gain > 1.0)
+            {
+                fprintf(stderr, "Invalid gain: %s (min: 0.0, max 1.0)\n", argv[i]);
+                tone_freq = 1;
```
kcat, line 261: Looks like a copy-paste error.

examples/altonegen.c
```diff
@@ -175,6 +181,7 @@ int main(int argc, char *argv[])
     ALint tone_freq = 1000;
     ALCint dev_rate;
     ALenum state;
+	ALfloat gain = 1.0f;
```
kcat, line 184: Doesn't seem indented properly.

examples/altonegen.c
```diff
@@ -144,6 +146,10 @@ static ALuint CreateWave(enum WaveType type, ALuint freq, ALuint srate)
             break;
     }
 
+	if (gain != 1.0)
+		for(i = 0;i < srate;i++)
+			data[i] *= gain;
```
kcat, line 151: This also seems over-indented.

examples/altonegen.c
```diff
@@ -44,7 +44,9 @@
 
 #include "common/alhelpers.h"
 
-#include "win_main_utf8.h"
+#ifdef _WIN32
+# include "win_main_utf8.h"
+#endif
```
kcat, line 49: Is there a reason this was put behind an `ifdef`, when the header already checks for Windows?

eliotmiranda: I made a mistake compiling the examples, leaving out an include directory path.  But for me I prefer having the logic that include or excludes platform specific files where I can see it rather than going to the header to check. So my steps went something like, “compilation on Mac failed, why? Saw the Windows specific include, ifdef’ed it out, compilation succeeded. Link failed with code in alhelpers.c missing.  Found alhelpers, added the directory, compiled and linked altonegen.  Didn’t go back to undo ifdefs cuz the code feels more readable that way.

I shall revert if if you prefer.

PR #504, opened 2020-12-11
Close PCM for proper update

TarcioV:
I have one problem while running alhrtf example in my Raspberry PI 2 wich use ALSA backend:

```
[ALSOFT] (WW) Failed to initialize backend "pulse"
Opened "ALSA Default"
AL_EXT_STEREO_ANGLES found
Available HRTFs:
    0: default-44100
    1: default-48000
    2: Built-In HRTF
Using default HRTF...
[ALSOFT] (WW) Error generated on device 0xadf460, code 0xa001
Failed to reset device: Invalid Device
HRTF not enabled!
```

Further investigation, guide me to this line of code:
https://github.com/kcat/openal-soft/blob/bd08b9a5b52f4f6f056d428399f92df92f82c00a/alc/backends/alsa.cpp#L766

We hit this line of code two times, one called by InitAL() and the other called by alcResetDeviceSOFT(), what happens is that on the second time the PCM is running wich according the ALSA documentation is a error (https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga1ca0dc120a484965e26cabf966502330):

`The hardware parameters cannot be changed when the stream is running (active)...`

This happens even after alcResetDeviceSOFT() calls AlsaPlayback::stop() in order to put the device in a state that can be reseted.

I can't find a function of ALSA that can put a running device in such state, so my solution was close the PCM and reopen it, so lsaPlayback::reset() can proper initialize it.

### DIFF
--- a/alc/backends/alsa.cpp
+++ b/alc/backends/alsa.cpp
@@ -835,11 +835,24 @@ void AlsaPlayback::start()
 
 void AlsaPlayback::stop()
 {
+    char *name{};
+
     if(mKillNow.exchange(true, std::memory_order_acq_rel) || !mThread.joinable())
         return;
     mThread.join();
 
     mBuffer.clear();
+
+    if(mPcmHandle)
+    {
+	name = strdup(snd_pcm_name(mPcmHandle));
+	snd_pcm_nonblock(mPcmHandle, 0);
+        snd_pcm_drain(mPcmHandle);
+	snd_pcm_nonblock(mPcmHandle, 1);
+        snd_pcm_close(mPcmHandle);
+	snd_pcm_open(&mPcmHandle, name, SND_PCM_STREAM_PLAYBACK, SND_PCM_NONBLOCK);
+	free(name);
+    }
 }
 
 ClockLatency AlsaPlayback::getClockLatency()


kcat:
> This happens even after alcResetDeviceSOFT() calls AlsaPlayback::stop() in order to put the device in a state that can be reseted.

This is likely because `AlsaPlayback::stop()` just stops the mixer processing, with the device being left to play what it has and then underrun (which a future `AlsaPlayback::play()` call will restart the mixer processing, which will detect the device underrun and fix it when resuming). I guess what's happening is either an underrun device remains 'active', or `snd_pcm_hw_params` is getting called too soon while the device still has samples.

Making `AlsaPlayback::stop()` call `snd_pcm_drop` and/or `snd_pcm_reset` after the mixer thread is finished may be the more appropriate solution. `snd_pcm_drain` might also be a better option, though I hear it's buggy and the docs aren't very clear about its behavior (does it wait until the device stops before returning? if not, do the other relevant functions know to wait on a draining device?).

kcat:
Commit bb597546e1db888120b0da5c23dbf249a281e6aa adds a `snd_pcm_drop` call when stopping the backend, rather than reopening the PCM device. Does that also fix the issue?

TarcioV:
Yes, it fixes the issue. I have tried to only use snd_pcm_drain() before, but it seams to not be enough to stop the PCM.
Thank you.

PR #511, opened 2020-12-25
fast2fi/fast_roundf implementations in asm for ARM64.

devnexen:


common/alnumeric.h
```diff
@@ -236,6 +236,10 @@ inline int fastf2i(float f) noexcept
 #endif
     return i;
 
+#elif (defined(__GNUC__) || defined(__clang__)) && defined(__aarch64__)
+    int i;
+    __asm__ volatile("fcvtzs %w0, %s1" : "=r"(i) : "w"(f));
+    return i;
```
kcat, line 242: I wonder if this is needed, and the standard cast to int doesn't automatically generate this assembly (or better, auto-vectorizing it with appropriate targets).

devnexen: Not always indeed I checked the assembly but good point.

PR #515, opened 2021-01-09
Pass nullptr to pulse for default device

### LINKED ISSUE #457: Distorted audio when recording stack is initialized
How to reproduce:

- Have an USB audio interface connected on Linux
- Have any Alsa or Pulseaudio program playing back music with a non-default playback rate (in my case 48000 is default, and Spotify plays back at 44100 Hz, which my USB audio interface supports as well so it plays back fine)
- Initialize the recording stack of OpenAL (see example down below)
- Now the playback of audio is distorted

Minimal reproducable OpenAL program:
```C
#include <AL/alc.h>

int main(int argc, char **argv)
{
	alcGetString(0, ALC_CAPTURE_DEFAULT_DEVICE_SPECIFIER);
	return 0;
}
```

Attached is the same program, compile with `gcc -l openal distorted_audio.c` and run it to cause the distorted audio.
[distorted_audio.zip](https://github.com/kcat/openal-soft/files/4978553/distorted_audio.zip)

I suspect that maybe OpenAL tries to open the recording device at 48000 Hz, while the same USB device is playing back at 44100 Hz, which would maybe cause issues?

This may however just be an Alsa bug and not a bug in this library, perhaps you can elaborate on what the initialization code may be doing and whether it could clash with any existing playback with different sample rates.

This originated from this issue: https://github.com/telegramdesktop/tdesktop/issues/1548
Telegram is basically executing the above code in its initialization routine.

But I can reproduce this bug with pure Alsa commands too, so I have already made a kernel bug report: https://bugzilla.kernel.org/show_bug.cgi?id=207617

Perhaps you know more about this matter? Should it even cause any problems from the Audio interface point of view, if you playback and record with different sample rates simultaneously?

ilya-fedin:
Pass nullptr to pulse when enumerating devices and in other in order to avoid device open, fix #457 and allow pulse to move stream to new default device when it changes (look https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/14.0/#significantroutingchangestodefaultsinkssources)

### DIFF
--- a/alc/backends/pulseaudio.cpp
+++ b/alc/backends/pulseaudio.cpp
@@ -611,24 +611,9 @@ void PulseMainloop::probePlaybackDevices()
         std::unique_lock<std::mutex> plock{mMutex};
 
         context = connectContext(plock);
-
-        constexpr pa_stream_flags_t flags{PA_STREAM_FIX_FORMAT | PA_STREAM_FIX_RATE |
-            PA_STREAM_FIX_CHANNELS | PA_STREAM_DONT_MOVE | PA_STREAM_START_CORKED};
-
-        pa_sample_spec spec{};
-        spec.format = PA_SAMPLE_S16NE;
-        spec.rate = 44100;
-        spec.channels = 2;
-
-        stream = connectStream(nullptr, plock, context, flags, nullptr, &spec, nullptr,
-            BackendType::Playback);
         pa_operation *op{pa_context_get_sink_info_by_name(context,
-            pa_stream_get_device_name(stream), &deviceSinkCallbackC, this)};
+            nullptr, &deviceSinkCallbackC, this)};
         waitForOperation(op, plock);
-
-        pa_stream_disconnect(stream);
-        pa_stream_unref(stream);
-        stream = nullptr;
 
         op = pa_context_get_sink_info_list(context, &deviceSinkCallbackC, this);
         waitForOperation(op, plock);
@@ -653,24 +638,9 @@ void PulseMainloop::probeCaptureDevices()
         std::unique_lock<std::mutex> plock{mMutex};
 
         context = connectContext(plock);
-
-        constexpr pa_stream_flags_t flags{PA_STREAM_FIX_FORMAT | PA_STREAM_FIX_RATE |
-            PA_STREAM_FIX_CHANNELS | PA_STREAM_DONT_MOVE | PA_STREAM_START_CORKED};
-
-        pa_sample_spec spec{};
-        spec.format = PA_SAMPLE_S16NE;
-        spec.rate = 44100;
-        spec.channels = 1;
-
-        stream = connectStream(nullptr, plock, context, flags, nullptr, &spec, nullptr,
-            BackendType::Capture);
         pa_operation *op{pa_context_get_source_info_by_name(context,
-            pa_stream_get_device_name(stream), &deviceSourceCallbackC, this)};
+            nullptr, &deviceSourceCallbackC, this)};
         waitForOperation(op, plock);
-
-        pa_stream_disconnect(stream);
-        pa_stream_unref(stream);
-        stream = nullptr;
 
         op = pa_context_get_source_info_list(context, &deviceSourceCallbackC, this);
         waitForOperation(op, plock);
@@ -726,7 +696,7 @@ struct PulsePlayback final : public BackendBase {
 
     PulseMainloop mMainloop;
 
-    std::string mDeviceName;
+    al::optional<std::string> mDeviceName{al::nullopt};
 
     pa_buffer_attr mAttr;
     pa_sample_spec mSpec;
@@ -852,7 +822,7 @@ void PulsePlayback::sinkNameCallback(pa_context*, const pa_sink_info *info, int 
 void PulsePlayback::streamMovedCallback(pa_stream *stream) noexcept
 {
     mDeviceName = pa_stream_get_device_name(stream);
-    TRACE("Stream moved to %s\n", mDeviceName.c_str());
+    TRACE("Stream moved to %s\n", mDeviceName->c_str());
 }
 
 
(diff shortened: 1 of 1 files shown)


alc/backends/pulseaudio.cpp
```diff
@@ -897,13 +857,12 @@ void PulsePlayback::open(const char *name)
     mStream = mMainloop.connectStream(pulse_name, plock, mContext, flags, nullptr, &spec, nullptr,
         BackendType::Playback);
 
-    pa_stream_set_moved_callback(mStream, &PulsePlayback::streamMovedCallbackC, this);
```
kcat, line 900: This is needed so that a stream that gets moved will stay on the device it moved to after a reset, instead of moving back to the old one.

alc/backends/pulseaudio.cpp
```diff
@@ -783,6 +783,30 @@ void PulsePlayback::sinkInfoCallback(pa_context*, const pa_sink_info *info, int
...
+        case PA_SAMPLE_FLOAT32NE:
+            mDevice->FmtType = DevFmtFloat;
+            break;
+        default:
+            break;
+        }
+    }
+
+    if(!mDevice->Flags.get<FrequencyRequest>())
+        mDevice->Frequency = info->sample_spec.rate;
```
kcat, line 810: The `PA_STREAM_FIX_RATE` flag should already handle this (and properly resize the buffer, as well). I'm also not sure about setting the sample type because the sink device will typically be `s16le`, but the server-side mixing, volume control, and any other module processing could benefit from having floating-point samples prior to 16-bit quantization.

alc/backends/pulseaudio.cpp
```diff
@@ -927,9 +934,13 @@ bool PulsePlayback::reset()
         flags &= ~PA_STREAM_EARLY_REQUESTS;
         flags |= PA_STREAM_ADJUST_LATENCY;
     }
+    if(!mDevice->Flags.get<SampleTypeRequest>() && !mSettingFormatFailed)
+        flags |= PA_STREAM_FIX_FORMAT;
     if(GetConfigValueBool(mDevice->DeviceName.c_str(), "pulse", "fix-rate", 0) ||
        !mDevice->Flags.get<FrequencyRequest>())
         flags |= PA_STREAM_FIX_RATE;
+    if(!mDevice->Flags.get<ChannelsRequest>() && !mSettingChannelsFailed)
+        flags |= PA_STREAM_FIX_CHANNELS;
```
kcat, line 948: This will cause the mixer to write incorrect channels (different order, too many, or too few) if it ends up getting changed by the server. Similar for the sample format, writing samples that are too large, too small, different signedness, etc, if it gets changed.

alc/backends/pulseaudio.cpp
```diff
@@ -891,14 +904,6 @@ void PulsePlayback::open(const char *name)
...
 
     mDeviceName = pulse_name ? al::make_optional<std::string>(pulse_name) : al::nullopt;
-    if(!dev_name)
-    {
-        pa_operation *op{pa_context_get_sink_info_by_name(mContext, pulse_name,
-            &PulsePlayback::sinkNameCallbackC, this)};
-        mMainloop.waitForOperation(op, plock);
-    }
-    else
-        mDevice->DeviceName = dev_name;
```
kcat, line 911: If there's multiple devices with the same description, OpenAL Soft appends `#2`, `#3`, etc, to them so they can be distinguished for the app to select between them. That needs to be maintained so the app sees the same device name that it used to open it, in case it wants to save/open the same device again.

ilya-fedin: I moved it to sinkInfoCallback since it's called everytime anyway now:
https://github.com/kcat/openal-soft/pull/515/commits/013876554952153cd79f7b656814745c981192ac#diff-c73de3dc72d31b51bc4950a6d02421036e422f9707eaf72c55fdf0caf5bffc43R784

alc/backends/pulseaudio.cpp
ilya-fedin: Isn't it expected to reset all moves on reset call?

alc/backends/pulseaudio.cpp
ilya-fedin: Ok, I'll remove that...

alc/backends/pulseaudio.cpp
ilya-fedin: How channels change should be handled? Should some buffer attributes be changed?

alc/backends/pulseaudio.cpp
kcat: That gets the description provided by PulseAudio, but like I said, if there happens to be multiple devices with the same description, OpenAL Soft appends `#2`, `#3`, etc, as necessary. For example, if I have two USB audio devices, and PulseAudio gives the description as `"USB Audio Analog Stereo"` for both of them, OpenAL Soft will enumerate them as:
```
USB Audio Analog Stereo
USB Audio Analog Stereo #2
```
so the desired one can still be selected/opened. If the app opens `"USB Audio Analog Stereo #2"`, the device name shouldn't be reported as `"USB Audio Analog Stereo"`. The `sinkInfoCallback` method just gets the description from PulseAudio, which doesn't account for this.

alc/backends/pulseaudio.cpp
kcat: No, `reset` changes/updates the internal playback parameters for the current device with what the app requests. If the playback stream was moved by the server, it should persist through resets. For example, if an app launches and starts playing through speakers by default, the user then uses pavucontrol or kmix or something to move the app playback to headphones, then the app reconfigures/resets the device to enable HRTF, it should keep playing on the headphones, not move back to the speakers.

alc/backends/pulseaudio.cpp
ilya-fedin: Got it

ilya-fedin:
Well, I removed almost all changes I did... :D

kcat:
Fails to build:
```
alc/backends/pulseaudio.cpp:825:46: error: no member named 'c_str' in 'al::optional<std::__cxx11::basic_string<char>>'; did you mean to use '->' instead of '.'?
    TRACE("Stream moved to %s\n", mDeviceName.c_str());
                                             ^
                                             ->
...
alc/backends/pulseaudio.cpp:1181:46: error: no member named 'c_str' in 'al::optional<std::__cxx11::basic_string<char>>'; did you mean to use '->' instead of '.'?
    TRACE("Stream moved to %s\n", mDeviceName.c_str());
                                             ^
                                             ->
```

ilya-fedin:
Oops, fixed

PR #543, opened 2021-03-18
Don't force application name with pulseaudio

ilya-fedin:
libpulse can read application name from an [environment variable](https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/Clients/ApplicationProperties/) and [fallbacks to binary name by default](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/proplist-util.c#L203-208). However, openal forces binary name as application name and makes impossible to set it from application :(

### DIFF
--- a/alc/backends/pulseaudio.cpp
+++ b/alc/backends/pulseaudio.cpp
@@ -38,7 +38,6 @@
 #include "alcmain.h"
 #include "alu.h"
 #include "alconfig.h"
-#include "compat.h"
 #include "core/logging.h"
 #include "dynload.h"
 #include "strutils.h"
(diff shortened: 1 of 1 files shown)


ilya-fedin:
@kcat can I ask, when the next release will be? Looking by older releases, it will be in the next year, right? :thinking:

kcat:
I honestly don't know. I prefer to make releases when I feel there's been enough good additions and/or important fixes. But it's not a bad idea to have more frequent releases as long as there's been some fixes, too.

ratijas:
Asked for a backport to Arch Linux. Let's see how it goes: https://bugs.archlinux.org/task/72729

PR #546, opened 2021-03-24
better conversion for ogg files with Libsndfile in examples/alplay.c

SeanTolstoyevski:
* for some ogg files, Libsndfile causes problems with float-int conversion. When playing the sound with Openal, treble and high DB points crack.
 - this issue is very old. But I checked it and it still seems to be a problem. https://github.com/SFML/SFML/issues/271
 - after the added code, the cracks were fixed

This is my **first pull request**. I'm learning.
Please comment for missing or problem.


kcat:
This attenuates the sound by as much as it exceeds the normalized range, as well as an extra `32768/32767`. I don't think I really like that, especially for code that's intended to be an example of how to do things; an unknown amount of volume reduction because of a few transient samples somewhere. A better idea may be to use OpenAL's floating-point formats, with the `AL_EXT_FLOAT32` extension, when the audio file format is more than 16-bit integer, though that does end up making it a bit more complex. It would be nice if libsndfile fixed this bug on their end, but they explicitly refuse to.

SeanTolstoyevski:
I think Ogg is a common audio format.
And libsndfile's too.

I can configure it according to the `float32` enums.

PR #559, opened 2021-04-30
Allow enabling SSE without SSE2

tatokis:
Some old CPUs such as Coppermine Pentium IIIs support SSE only, and Linux distributions like ArchLinux32 have separate architectures and repositories for SSE and SSE2 CPUs.

This patch allows individual control over SSE and SSE2 support.
It has been tested on a machine that only supports SSE.

I'm not sure why mixer_sse.cpp forced SSE2, as all the functions in there are SSE.


kcat:
Is there much of an improvement with this? I'd figure the performance penalty of using x87 FPU math (specifically the denormals that can easily get introduced in various processing loops) would be really bad, especially on slower SSE1 CPUs, and that leveraging the few SSE1 routines wouldn't do much to get good performance. Perhaps if you avoid using filters and effects with feedback loops, I suppose, some basic output could work without running into them.

> I'm not sure why mixer_sse.cpp forced SSE2, as all the functions in there are SSE.

Mainly because since SSE support also requires SSE2 anyway, enabling an SSE2 target could allow the compiler to optimize with any relevant SSE2 instructions.

tatokis:
Apologies for the late response and thanks for merging it.

I did run some very quick tests.
Just running the compiler with `-msse` in the library results in a ~20% improvement in this https://github.com/ahicks92/audio_benchmarks/blob/master/benchmarks/OpenAL/main.cpp
For something more complex, I tried checking how much CPU time the `almultireverb` example took, and it was within margin of error, but tended to be slightly less.

So overall from some very simple testing, it was either the same performance or better.

If you have a specific test you'd want me to run, I'd be happy to, but unfortunately I don't have the time to write a proper benchmark.

kcat:
I meant more in general when using x87 math. Having more than a few sources with filters, and a few effects like reverb, would cause a lot of denormals to be generated in their feedback loops, which would bring performance down, and enabling some SSE routines wouldn't help much with making it usable as long as x87 codegen was the primary target.

PR #562, opened 2021-05-12
Add android studio gradle project

halx99:


sgowen:
Hmmm, so this was never merged? I'd like to use OpenAL-soft on Android if possible.

halx99:
it's only for bug reproduce

sgowen:
Ah I see. I got excited. An example Android Studio project demonstrating building and running openal-soft would be sweet.

kcat:
OpenAL Soft is a library for playing 3D audio. The library itself doesn't run, it needs to be added to an app/game. There are a few example programs it comes with, but they're very simple general programs, which are built as part of the CMake project. The library would be added to an existing project, so it having its own project probably wouldn't be useful. But either way, OpenAL Soft is a rather standard CMake project, so you can use it the same way you'd use a CMake project on Android.

PR #565, opened 2021-05-14
Improve cmake with option 'ALSOFT_EXAMPLES'

halx99:


kcat:
What makes this better? The library is tagged with `EXCLUDE_FROM_ALL` so it won't be built unless another target that does get built needs it. This change can pose a problem if the library ever gets used by non-example targets.

halx99:
You are right.

Issue #566, opened 2021-05-14
Compiling Clang in Windows get issues with min and max


Togtja:
So I seem to be suffering from the issue outline in this stack overflow [question ](https://stackoverflow.com/questions/11544073/how-do-i-deal-with-the-max-macro-in-windows-h-colliding-with-max-in-std)

Where ever you use `std::numeric_limits<int>::max()` or `std::max()` the Windows API believes you to be using it's marco for `min/max` the solution is to define `NOMINMAX` or change all the `std` to have parentheses around it like this: `(std::numeric_limits<int>::max)()`

kcat:
Where is the error happening? `NOMINMAX` s defined here:
https://github.com/kcat/openal-soft/blob/1d7ff54f/CMakeLists.txt#L208
and applied to the library here:
https://github.com/kcat/openal-soft/blob/1d7ff54f/CMakeLists.txt#L1369

Togtja:
Alot of places. So I use the newest Clang (12.0.0) compiling openal-soft's master. I first get stopped on:
```
[build] ../../../common/threads.cpp:67:74: error: too few arguments provided to function-like macro invocation
[build]     if(initial > static_cast<unsigned int>(std::numeric_limits<int>::max()))
....
[build] ../../../common/threads.cpp:69:76: error: too few arguments provided to function-like macro invocation
[build]     mSem = CreateSemaphore(nullptr, initial, std::numeric_limits<int>::max(), nullptr);
````
Then after solving those. I get stopped on:
```
[build] ../../../common/almalloc.h:91:54: error: too few arguments provided to function-like macro invocation
[build]         if(n > std::numeric_limits<std::size_t>::max()/sizeof(T)) throw std::bad_alloc();
```
Then I get stopped:
```
[build] ../../../alc/backends/wasapi.cpp:129:76: error: too few arguments provided to function-like macro invocation
[build]     return static_cast<uint>(mini64(retval, std::numeric_limits<uint>::max()));
...
[build] ../../..\al/source.h:49:55: error: too few arguments provided to function-like macro invocation
[build]     float MaxDistance{std::numeric_limits<float>::max()};
...
[build] ../../../alc/alc.cpp:2665:47: error: too few arguments provided to function-like macro invocation
[build]                 std::numeric_limits<int>::max()));
```
And it continues like this, if you want the rest, please let me know, and I can go through them

Togtja:
[...]

PR #567, opened 2021-05-14
Fix windowsAPI issue with min/max

Togtja:
Fixes the issue with clang not compiling due to the WinAPI having defined min/max #566 
The other simpler solution is to define `NOMINMAX` but don't know where to start that, so just did what I could

### DIFF
--- a/al/buffer.cpp
+++ b/al/buffer.cpp
@@ -537,7 +537,7 @@ void LoadData(ALCcontext *context, ALbuffer *ALBuf, ALsizei freq, ALuint size,
             "Data size %d is not a multiple of frame size %d (%d unpack alignment)",
             size, SrcByteAlign, align);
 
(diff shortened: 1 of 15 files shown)


kcat:
> The other simpler solution is to define NOMINMAX but don't know where to start that, so just did what I could

It's currently defined [here for MSVC](https://github.com/kcat/openal-soft/blob/4cc820b/CMakeLists.txt#L213). Does Clang want to be treated as MSVC or GCC/MinGW on Windows? If it's a macro that should be set for all Windows compilers instead of just MSVC, it can be moved to [line 124](https://github.com/kcat/openal-soft/blob/4cc820b/CMakeLists.txt#L124) instead,

Togtja:
It's a while is I looked at it. But I believe Clang want to have the `NOMINMAX` set same as MSVC. So I think that would be the better solution

kcat:
Commit 602b33fede0ce7d158ca96cda6db45742b2b2fce should handle this more cleanly now, if you can check.

PR #578, opened 2021-06-21
Fix getting of device channel count in CoreAudio backend

alexey-lysiuk:
AudioBufferList contains a variable length array of mNumberBuffers elements, so it should not be created with the default constructor like usual class instances.
Unfortunately, Apple developer site documentation is literally empty for this API. There is a bunch of comments in framework's header files.
Here is the correct usage of AudioBufferList pointer from Chromium: https://chromium.googlesource.com/chromium/src/media/+/008a1abc573e1f8bcf513b50ce48a923b30ef130/audio/mac/audio_manager_mac.cc#266
There were occasional crashes because of memory corruption which was confirmed by address sanitizer

### DIFF
--- a/alc/backends/coreaudio.cpp
+++ b/alc/backends/coreaudio.cpp
@@ -141,14 +141,13 @@ UInt32 GetDeviceChannelCount(AudioDeviceID devId, bool isCapture)
     }
 
     auto buffer_data = std::vector<char>(propSize);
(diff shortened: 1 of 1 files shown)


kcat:
If it doesn't need to be `new`'d/constructed, it probably shouldn't be destroyed/destructed either (with the `al::destroy_at` calls).

alexey-lysiuk:
Yep, `AudioBufferList` destructor does nothing anyway.

PR #581, opened 2021-07-05
Make OpenALConfig.cmake compatible with CMake's FindOpenAL.cmake

madebr:
Hello,

The current `OpenALConfig.cmake` script, created + installed by the cmake build script is not compatible with CMake's official [`FindOpenAL.cmake`](https://cmake.org/cmake/help/latest/module/FindOpenAL.html):
- The variables `OPENAL_INCLUDE_DIR`, `OPENAL_LIBRARY` and `OPENAL_VERSION_STRING` are not defined.
- `#include "al.h"` is not possible.

This pr fixes this by using cmake generator expressions to automatically export these variables from the installed `OpenAL::OpenAL` target.
A `OPENAL_DEFINITIONS` variable is also introduced to provide eventual compile definitions.

The `AL` subfolder is also added as include directory to enable including `"al.h"`.

details:
- I had to add `cmake_minimum_required(VERSION 3.1)` because the generator expression `LINK_ONLY` is available since cmake 3.1.
- Having included [`CMakePackageConfigHelpers`](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html) to create `OpenALConfig.cmake`, I used the opportunity to also create a `OpenALConfigVersion.cmake` script.
This allows users of OpenAL to require a certain minimum version using e.g. `find_package(OpenAL 1.21.0 REQUIRED CONFIG)`.

**Why do I need this fix?**

I [encountered a cmake toolchain](https://github.com/devkitPro/pacman-packages/commit/d9f8148c77c09023e1bffb333ef125c40307e8c8) that sets `CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`, this caused cmake to use openal-soft's cmake script instead of the one from cmake which causes missing openal symbols when linking.
Granted, I can fix this by adding `MODULE` to `find_package(OpenAL REQUIRED)` but I think a lot of cmake build scripts don't do this.

**Local tests performed**
I only tested on Linux.
I created a small openal project and linked to openal-soft in both shared and static configuration.
I tested building with `OpenALConfig.cmake` in the build folder, and buidling with `OpenALConfig.cmake` in an install prefix.


kcat:
> This allows users of OpenAL to require a certain minimum version using e.g. `find_package(OpenAL 1.21.0 REQUIRED CONFIG)`.

I'm not sure about this. The OpenAL cmake module should be implementation-independent, whereas a version like `1.21.0` is OpenAL Soft specific. At compile time, the available API is essentially the 1.1 standard*, and extra functionality is determined by checking extensions and loading function pointers at run-time. Maybe if there was a way to specify the desired implementation (e.g. `find_package(OpenAL ALSOFT-1.21.0 REQUIRED)`), but I'm not sure that's the best way to handle it.

(*) With OpenAL Soft, an app can technically use certain extension functions directly if they have a hard requirement on the extension existing, but this is discouraged since something like the router (on Windows) can prevent direct access to OpenAL Soft functions at link-time and/or load-time, even though the extension and functions can still be accessed dynamically.

madebr:
What you're proposing/asking here is a completely custom cmake version configuration file which would be considerably more work.
I'll drop the version related change as it was supposed to be an extra and not required to be `FindOpenAL.cmake`-compatible.

PR #607, opened 2021-10-22
Update alsoftrc.sample

fatboyaj:


kcat:
What is this for? Why is the sample config file setting a couple options at the end like that?

PR #618, opened 2021-11-20
Change user config path to match system-wide directory structure

etircopyh:


kcat:
The system-wide directories are in `XDG_CONFIG_DIRS` (or `/etc/xdg`) don't use the `openal` sub-directory. The `/etc/openal/alsoft.conf` config is legacy (along with `$HOME/.alsoftrc`), and it only uses a sub-directory to avoid dumping directly in `/etc`.

I don't know about this. Normally it's supposed to be `$XDG_CONFIG_HOME/appname/files`. But OpenAL isn't an app, it's a library used by multiple apps. Though even if libraries use a sub-directory like that, since there's only one config file, it doesn't seem necessary. Other apps that only have a single config file don't seem to use a sub-directory: `$XDG_CONFIG_HOME/kdeveloprc`, `$XDG_CONFIG_HOME/glade.conf`, `$XDG_CONFIG_HOME/pavucontrol.ini`, etc.

Anyone else have any input on this?

PR #632, opened 2022-01-02
Add EAX extensions (EAX 2.0-5.0, X-RAM)

bibendovsky:
CMake options:
- `ALSOFT_EAX` - compile EAX related code. Default is `ON`.

Configuration file:
- `eax/enable` - enables EAX extensions. Default is `true`.


bibendovsky:
Binaries to test.
[openal_soft_eax_bin_win32_pr632_98754aa.zip](https://github.com/kcat/openal-soft/files/7799563/openal_soft_eax_bin_win32_pr632_98754aa.zip)

mirh:
I feel like the patcher application could probably live on your own separate tree.

kcat:
I'd rather not have the patcher itself as part of the library. It would be far more preferable to implement the extensions properly, instead of a built-in process/executable patcher and wrapper.

bibendovsky:
Removed patch related stuff.

[openal_soft_eax_bin_win32_pr632_a01d1dc.zip](https://github.com/kcat/openal-soft/files/7802362/openal_soft_eax_bin_win32_pr632_a01d1dc.zip)

kcat:
This will need a lot of work to integrate properly. There's a lot of unnecessary code and wrapping behavior that in-built extensions don't need. It has direct access to the implementation internals, so there's no need to track the current context separately for EAX (since the code can be given the current context handle), there's no need for wrapper calls through virtual interfaces when the code can be added right into the calls themselves, no need for wrapper objects (since the information can be stored in the implementation objects, if it's not already there), etc.

Kappa971:
With this change, if I understand correctly, EAXEFX would no longer be needed for games like Doom 3 as EAX would be natively supported by OpenAL Soft, right? Will this also lead to changes to DSOAL?

mirh:
Doom 3 would still need the no shaking patch AFAICT. 
And while DSOAL could possibly be moved to an alchemy-like architecture, I guess the final deal if any is actually the opposite to address some of its [limitations](https://github.com/kcat/dsoal/issues/32#issuecomment-971695436) (i.e. have openal self-contained in dsound.dll)

Kappa971:
> Doom 3 would still need the no shaking patch AFAICT. And while DSOAL could possibly be moved to an alchemy-like architecture, I guess the final deal if any is actually the opposite to address some of its [limitations](https://github.com/kcat/dsoal/issues/32#issuecomment-971695436) (i.e. have openal self-contained in dsound.dll)

All this would be a further step towards "EAX for all", as well as saying goodbye to Creative and its policy against its own users (Creative has completely abandoned the support of ALchemy and Host OpenAL and has no intention of releasing any source codes or documentation) . 
We can only thank @kcat and @bibendovsky for their work, I hope this will be implemented soon.

bibendovsky:
Rewrote the integration.
Changed the name of option `eax/disable` to `eax/enable`. Default value is `true`.

[openal_soft_eax_bin_win32_pr632_e57c73a.zip](https://github.com/kcat/openal-soft/files/7920955/openal_soft_eax_bin_win32_pr632_e57c73a.zip)

juliuscaesarnl:
As one of the end users of this project, without knowing much about the implementation, I can confirm that EAX 5.0 works when playing BioShock with this build. Keep up the good work @bibendovsky 👍

al/auxeffectslot.cpp
```diff
@@ -1042,3 +1046,840 @@ EffectSlotSubList::~EffectSlotSubList()
...
+}
+
+void ALeffectslot::eax_set_fx_slot_effect()
+{
+    if (false)
+    {
+    }
+    else if (eax_eax_fx_slot_.guidLoadEffect == EAX_NULL_GUID)
+    {
+        eax_set_fx_slot_effect(AL_EFFECT_NULL, eax_null_effect_);
```
kcat, line 1471: Is it intended that when an fxslot changes its effect type (e.g. chorus to reverb, or reverb to null), the parameter values for the previous effect type are retained for when it's changed back?

bibendovsky: It's not. I'll fix that.

kcat: It should be a fairly easy fix. Rather than a bunch of `EaxAlEffectUPtr eax_<type>_effect_{};` with one `ALeffect* eax_al_effect_{};` that points to the "current" one, it just needs one `eax_al_effect_`, and changing the effect type sets all properties for that type to defaults (which then get applied on the effect object as normal).

kcat:
Well, there are still some things that could do with cleaning up, but they're relatively minor which I can easily fix up. And there's nothing jumping out at me as particularly wrong. So given no reports of it being a problem, I guess it's time to get this merged in and see how it goes. Thanks!

PR #653, opened 2022-02-01
[EAX] Add primary extension name fro EAX v2.0

bibendovsky:
Seems the extension name `EAX2.0` was introduced later.
The manual uses name `EAX` to query for the extension presence.


mirh:
Well, no I guess then
https://github.com/kcat/dsoal/issues/43#issuecomment-1033102798

PR #693, opened 2022-04-27
Search the installation data directory

ilya-fedin:
That allows the logic to work on non-FHS distros like NixOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,11 @@ option(ALSOFT_UPDATE_BUILD_VERSION "Update git build version info" ON)
 
 option(ALSOFT_EAX "Enable legacy EAX extensions" ${WIN32})
 
+option(ALSOFT_SEARCH_INSTALL_DATADIR "Search the installation data directory" OFF)
+if(ALSOFT_SEARCH_INSTALL_DATADIR)
+    set(ALSOFT_INSTALL_DATADIR ${CMAKE_INSTALL_FULL_DATADIR})
+endif()
+
 if(DEFINED SHARE_INSTALL_DIR)
     message(WARNING "SHARE_INSTALL_DIR is deprecated.  Use the variables provided by the GNUInstallDirs module instead")
     set(CMAKE_INSTALL_DATADIR "${SHARE_INSTALL_DIR}")
--- a/config.h.in
+++ b/config.h.in
@@ -114,3 +114,6 @@
 
 /* Define if we have pthread_set_name_np() */
 #cmakedefine HAVE_PTHREAD_SET_NAME_NP
+
+/* Define the installation data directory */
+#cmakedefine ALSOFT_INSTALL_DATADIR "@ALSOFT_INSTALL_DATADIR@"
--- a/core/helpers.cpp
+++ b/core/helpers.cpp
@@ -408,6 +408,20 @@ al::vector<std::string> SearchDataFiles(const char *ext, const char *subdir)
         DirectorySearch(path.c_str(), ext, &results);
     }
 
+#ifdef ALSOFT_INSTALL_DATADIR
+    // Search the installation data directory
+    {
+        std::string path{ALSOFT_INSTALL_DATADIR};
+        if(!path.empty())
+        {
+            if(path.back() != '/')
+                path += '/';
+            path += subdir;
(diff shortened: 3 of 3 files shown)


ilya-fedin:
@kcat ping

kcat:
I'm not sure I like encoding the install path into the library like that, because it will then misbehave if it's moved elsewhere. It's not uncommon for a package to be built and installed to a custom directory which gets packaged up, then installed on the user's system in a different directory. Or to be built and installed into whatever directory, then packaged as part of a self-contained app. And especially if the install directory is the same as one that's already searched, causing duplicate results.

Is there a way to get the relevant data directories at runtime if the XDG standard isn't good enough?

ilya-fedin:
> It's not uncommon for a package to be built and installed to a custom directory which gets packaged up, then installed on the user's system in a different directory.

Yeah, it's helpful only when building for system installation. Maybe openal has some cmake variable to check whether the library should be relocatable?

> Is there a way to get the relevant data directories at runtime if the XDG standard isn't good enough?

Only if we can get the location of the library...

kcat:
> Yeah, it's helpful only when building for system installation. Maybe openal has some cmake variable to check whether the library should be relocatable?

Not currently, but an option to hardcode the install data path could be added. Something like
```
option(ALSOFT_SEARCH_INSTALL_DATADIR ...)
if(ALSOFT_SEARCH_INSTALL_DATADIR)
    set(ALSOFT_INSTALL_DATADIR ${CMAKE_INSTALL_FULL_DATADIR})
endif()

...

#cmakedefine ALSOFT_INSTALL_DATADIR

...

#ifdef ALSOFT_INSTALL_DATADIR
// Search the installation data directory
{
    std::string path{ALSOFT_INSTALL_DATADIR};
    ...
```
Then set `-DALSOFT_SEARCH_INSTALL_DATADIR=TRUE` when running cmake, and it should include the installation data directory when searching.

ilya-fedin:
That seem to work:
```
    Dl_info libinfo;
    if(dladdr(reinterpret_cast<void*>(&SearchDataFiles), &libinfo))
    {
        std::string path{libinfo.dli_fname};
        if(!path.empty())
        {
            path = std::string{dirname(dirname(&path.front()))};
        }
        if(!path.empty())
        {
            path += "/share/";
            path += subdir;
            DirectorySearch(path.c_str(), ext, &results);
        }
    }
```
But would need separating backends' HAVE_DLFCN_H logic as this needs to work even when backends needs to be linked it

ilya-fedin:
What do you think is better?

kcat:
I think I'd prefer the option approach. Getting the library path with `dladdr` and getting the data path relative to it requires presuming the directory structure, and opens up potential issues when the library is moved around (and would also need to have extra checks that the path isn't a duplicate of another path already checked). The option method allows for the install data directory to be included at the user or builder's discretion, when they know it's the proper thing to do.

ilya-fedin:
What should be the default (ON/OFF)?

kcat:
OFF, since it's only relevant for particular build/install environments.

ilya-fedin:
btw, while building I noticed a lot of warnings from pipewire headers, you might want to do this
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 31027577..6900c0c8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1392,6 +1392,8 @@ else()
     endif()
 endif()
 
+target_include_directories(${IMPL_TARGET} SYSTEM PRIVATE ${INC_PATHS})
+
 target_include_directories(${IMPL_TARGET}
   PUBLIC
     $<BUILD_INTERFACE:${OpenAL_SOURCE_DIR}/include>
@@ -1400,7 +1402,6 @@ target_include_directories(${IMPL_TARGET}
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/AL>
   PRIVATE
-    ${INC_PATHS}
     ${OpenAL_BINARY_DIR}
     ${OpenAL_SOURCE_DIR}
     ${OpenAL_SOURCE_DIR}/common
```

ilya-fedin:
> OFF, since it's only relevant for particular build/install environments.

Done

PR #696, opened 2022-04-30
Fix pkg-config file when libdir/bindir/includedir specified as absolute paths

ilya-fedin:
Currently it makes the pkg-config file unusable as prefix is specified twice.

FULL variables do the necessary parsing and automatically prepend the prefix if needed.


ilya-fedin:
@kcat can you take a look on this, too?

PR #705, opened 2022-05-22
[EAX] Add separate effect state for each version

bibendovsky:
How it works
========

EAXGet
1. Validate input parameters and return the properties of the version specified in the call.
2. Update a version to use in commit.

EAXSet
1. Validate input parameters and update deferred properties of the version specified in the call.
2. Update a version to use in commit.

Commit
1. Use a version specified in the last EAXGet/EAXSet call to select a versioned state.
2. Save the current properties.
3. Overwrite a versioned immediate properties with the deferred ones.
4. Overwrite a current properties with a deferred versioned one.
5. Compare saved properties with a current one and update appropriate EFX properties.


al/auxeffectslot.h
```diff
@@ -91,6 +91,7 @@ struct ALeffectslot {
 
 private:
     ALCcontext* eax_al_context_{};
+    const EaxCall* eax_call_{};
```
kcat, line 94: I don't think I like this being something stored in the object, which is updated before an `eax_dispatch` call. It makes it harder to reason about what it's actually doing, since the information is filled in and stored elsewhere. Especially as a pointer to const, which indicates the object shouldn't be getting modified.

bibendovsky: Fixed.

PR #708, opened 2022-05-30
Upload Win32 and x64 artifacts to Github actions

ThreeDeeJay:
Provides an alternative Windows binary download location that expires in 3 months (2 months longer than Appveyor)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,17 @@ jobs:
       matrix:
         config:
         - {
-            name: "Visual Studio 64-bit",
+            name: "Win32",
+            os: windows-latest,
+            cmake_opts: "-A Win32 \
+              -DALSOFT_BUILD_ROUTER=ON \
+              -DALSOFT_REQUIRE_WINMM=ON \
+              -DALSOFT_REQUIRE_DSOUND=ON \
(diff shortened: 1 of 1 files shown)


kcat:
Won't this also try to create artifacts for Linux and macOS? I'm not the most familiar with github ci scripts, but it seems like it would.

ThreeDeeJay:
Ah, yes, forgot to mention. I'm not really familiar with Linux/macOS binaries so I'm not sure what file/s are needed for them.
So I just made it upload `build/Release/soft_oal.dll` which works (tested both 32/64-bit DLLs) on Windows, at least to have a bit more consistency with DSOAL's Actions while https://github.com/kcat/dsoal/issues/13 gets sorted out.
Alternatively, perhaps [Deledrius](https://github.com/Deledrius) might be able to complete the puzzle for the missing binaries.

kcat:
I'm not too worried about Linux or macOS having artifacts right now, I just want to be sure it's not going to be an issue for it to try packaging Windows artifacts during Linux or macOS builds.

ThreeDeeJay:
In that case, I just added a check to skip uploading non-windows artifacts. [Before ](https://github.com/ThreeDeeJay/openal-soft/actions/runs/2406653149)- [After](https://github.com/ThreeDeeJay/openal-soft/actions/runs/2411263508)

PR #729, opened 2022-07-10
Workaround when using MinGW x86 when compiling

### LINKED ISSUE #715: Failed to build openal-soft with mingw81_32
MinGW Version: 8.1 32 bit (from Qt 5.15.2)
OS: Win10
OpenAL-Soft Version: master-branch
CMake GUI from Qt 5.15.2

I set three check boxes:

image

image

quyykk:
This fixes #715. The current version of windres doesn't parse command line arguments correctly when there's a space between them. The part that causes the error is a `-D` directive that defines a macro with two attributes, separated by space. That space is the problem and causes windres to misparse the argument, even if the two attributes are quoted.

The workaround here is to simply remove the space between the two attributes. I can confirm that with this fix, GCC 12 + MinGW 10 x86 is able to successfully build openal-soft.


kcat:
I had a commit for this already, which I forgot to push; commit fa51c89549590319cb545a8c81419e2e1ddc5db3.

PR #730, opened 2022-07-11
[EAX] Use separate FX slot state for each version

bibendovsky:
Also removes the EAX deferring of FX slot properties.

### DIFF
--- a/al/auxeffectslot.cpp
+++ b/al/auxeffectslot.cpp
@@ -50,11 +50,6 @@
 #include "effect.h"
 #include "opthelpers.h"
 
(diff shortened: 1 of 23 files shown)


bibendovsky:
> I don't like the removal of deferring slot properties

Sorry for innacurate description. I mean _EAX deferring_. FX slot does not have special flags to commit/defer properties, like context, source or an effect (i.e. EAXCONTEXT_PARAMETER_DEFER, EAXSOURCE_PARAMETER_DEFERRED, EAXREVERB_DEFERRED, .etc).

PR #733, opened 2022-07-17
Fix FreeBSD build proposal.

devnexen:


core/helpers.cpp
```diff
@@ -497,8 +497,8 @@ bool SetRTPriorityRTKit(int prio)
         if(getrlimit(RLIMIT_RTTIME, &rlim) != 0)
             return errno;
 
-        TRACE("RTTime max: %llu (hard: %llu, soft: %llu)\n", umaxtime, ulonglong{rlim.rlim_max},
-            ulonglong{rlim.rlim_cur});
+        TRACE("RTTime max: %llu (hard: %llu, soft: %llu)\n", umaxtime, ulonglong{static_cast<ulonglong>(rlim.rlim_max)},
+            ulonglong{static_cast<ulonglong>(rlim.rlim_cur)});
```
kcat, line 501: It's redundant to do both a plain `ulonglong{...}` cast and a `static_cast<ulonglong>(...)`. But what type is `rlim_t` on FreeBSD? An `unsigned long long` is the largest standard type, so it should be guaranteed to fit.

devnexen: it s signed on freebsd.

kcat:
This is already handled now by commits 1b28a24f6e917380536d882384aba388e331329e and b26ca6b65e703fdc852e451945582f0ed6cc9c35.

PR #752, opened 2022-08-18
rtkit: remove syscall include

neheb:
Not present under MinGW.

### DIFF
--- a/core/rtkit.cpp
+++ b/core/rtkit.cpp
@@ -40,8 +40,9 @@
 #include <string.h>
 #include <unistd.h>
 #include <sys/types.h>
+#ifdef __linux__
 #include <sys/syscall.h>
-#ifdef __FreeBSD__
+#elif defined(__FreeBSD__)
 #include <sys/thr.h>
 #endif
 


neheb:
Issue I discovered while developing https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/openal-soft/meson.build

neheb:
wait... why is this not using std::this_thread::get_id ?

kcat:
> wait... why is this not using std::this_thread::get_id ?

Because that's not guaranteed to match the TID provided by the kernel, which is what needs to be sent to the RTKit service. `std::thread::id` is an opaque (if trivially copyable) class that can be sorted, and serialized with `operator<<` (for e.g. `std::cout`). But other than being unique for each thread, there's no guarantee about what the class's value is.

kcat:
> Issue I discovered while developing https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/openal-soft/meson.build

I would not recommend creating a new build script. There are nuances in what CMakeLists.txt checks for and sets up to make a proper build, and doing it incorrectly can create builds that seem to work, but contain potentially severe problems, such as stack misalignment for exported functions, or exporting functions that aren't supposed to be (creating ABI compatibility issues when updating). And as you're discovering here, including sources that shouldn't be included for a target (RTKit/DBus isn't used for Windows, so they should be skipped there).

neheb:
In any case, this PR is the correct solution. The includes match the code.

neheb:
as far as duplicating the build system, the situation is the following:

meson supports wrapping cmake builds just fine. https://github.com/libusb/hidapi/pull/410 for an example. Completely unacceptable for WrapDB.

One way I could do it is by having such a thing merged here and _hide_ that implementation detail in wrapdb by not providing a custom meson.build file. But that's probably unacceptable here.

There's no good solution.

kcat:
At the very least, you should build with `-fvisibility=hidden`, and set `__attribute__((visibility("protected")))` for `ALC_API` and `AL_API`, to prevent exporting everything (on targets that support it) and not allow other code to override internal references to exported functions.

I don't know how much you care about 32-bit x86, but SSE2 builds are made by default on 32-bit since it's good to avoid the x87 FPU (build with `-msse2 -mfpmath=sse`). 32-bit GCC builds should also have `__attribute__((force_align_arg_pointer))` on `ALC_API` and `AL_API`, since 32-bit Windows and older 32-bit Linux code don't guarantee a suitably aligned stack, while GCC and MinGW will assume it.

neheb:
with a 32bit mingw build, I see,

```
/* Define if we have SSE CPU extensions */
#define HAVE_SSE
#define HAVE_SSE2
#define HAVE_SSE3
#define HAVE_SSE4_1
```

The force_align_arg_pointer thing is interesting. Do you know if ARM32 is also impacted?

kcat:
> with a 32bit mingw build, I see,

Yes, that just means it can build the specialized SSE1/2/3/4 mixer/resampler functions, not that it's building with SSE codegen for non-specialized functions. `-msse2` is actually a requirement on 32-bit GCC for `HAVE_SSE_INTRINSICS`, since the latter allows the library to call the `_mm_*` intrinsics in plain code, which requires SSE codegen to be enabled by the former (and `-mfpmath=sse` tells GCC/Clang to use the SSE unit for all floating point math instead of the x87 FPU, which has performance implications). x86-64 uses SSE2 by default, so those command-line switches are unneeded with 64-bit, and MSVC also uses `/arch:SSE2` by default even on 32-bit, so it's just 32-bit GCC/Clang that needs `-msse2 -mfpmath=sse`).

> The force_align_arg_pointer thing is interesting. Do you know if ARM32 is also impacted?

I don't know for sure, but I doubt it. I don't think ARM is as sensitive to alignment as SSE is, and systems were already starting to use larger alignments for function calls as those CPUs became more prominent.

PR #766, opened 2022-09-08
dynload: switch conditions

neheb:
MinGW has a dlfcn wrapper than can be used.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/common/dynload.cpp
+++ b/common/dynload.cpp
@@ -5,21 +5,7 @@
 
 #include "strutils.h"
 
-#ifdef _WIN32
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-
-void *LoadLib(const char *name)
-{
-    std::wstring wname{utf8_to_wstr(name)};
-    return LoadLibraryW(wname.c_str());
-}
-void CloseLib(void *handle)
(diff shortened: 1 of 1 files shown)


kcat:
Is there some info about how MinGW implements dlopen/dlsym, etc? Particularly the encoding for the filename, since it takes a `const char*` which is presumed to be UTF-8 outside of Windows, but Windows typically treats char strings as encoded in the current codepage instead. I don't know if MinGW's dlopen wrapper treats it as ASCII and passes it to `LoadLibraryA`, which would be a problem, or treats it as UTF-8 and converts it to wide char/UTF-16/UCS-2 to pass to `LoadLibraryW`.

neheb:
code is here: https://github.com/dlfcn-win32/dlfcn-win32/blob/master/src/dlfcn.c

I see LoadLibraryExA being used.

kcat:
That would be a problem since OpenAL Soft is handling UTF-8 strings, not the current code page that `LoadLibraryExA` expects. But even if it was converted, a filename can include characters from different character sets that can't all be represented in the active code page.

It seems better to stick with `LoadLibraryW` here.

neheb:
Hrm OK

PR #771, opened 2022-09-23
Fixed bug, possible overflow type and non-void no return

GermanAizek:
hi @kcat,
I think you'll like these little changes in PR.


kcat:
The `source.h` change should be unnecessary since the `eax_fail_` function throws and is marked `[[noreturn]]`, making the return statement unreachable code. For `win_main_utf8.h`, there's two places that do that addition and cast.

GermanAizek:
@kcat fix it https://github.com/kcat/openal-soft/pull/771/commits/a6ea00332bd5ea1d2ee221774ad334b2de60c8df

PR #776, opened 2022-10-09
Provide an implementation for the PS4 SceAudio backend

nkrapivin:
This provides a mostly complete implementation of an AL backend for the PlayStation 4 platform.

It does not provide any build scripts or any project files, as this is up to game developers to manage.

I didn't add it into CMake because this is a closed console platform, there is no point for it to be in CMake, console devs usually have their own build systems to manage such things.

If there are any questions I can try to answer them, but this backend alone should do 90% of the work.

### DIFF
--- a/alc/alc.cpp
+++ b/alc/alc.cpp
@@ -151,6 +151,9 @@
 #ifdef HAVE_SDL2
 #include "backends/sdl2.h"
 #endif
+#ifdef HAVE_SCEAUDIO
+#include "backends/sceaudio.h"
+#endif
 #ifdef HAVE_WAVE
 #include "backends/wave.h"
 #endif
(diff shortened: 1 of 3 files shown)


kcat:
This doesn't look like a good idea. If the library can't even be built for the system without more changes, I don't think there should be a backend for that system. Comments like "not gonna explain any of these, just take these for granted" with bare function declarations make me concerned too, as it makes it sound like they shouldn't be publicized. And if there's no public documentation, or even a clean-room reverse-engineered SDK that would have information about the APIs, if there's an issue with the code, I wouldn't even know where to begin to try to fix it.

nkrapivin:
Yeah, good idea tbh, if anybody needs this they can clone my branch.

PR #783, opened 2022-11-13
Add GitHub actions workflow to build utils/makemhr on update

ThreeDeeJay:
This workflow [builds makemhr.exe including dependencies (libmysofa, zlib, etc)](https://github.com/ThreeDeeJay/openal-soft/actions/runs/3457267682/jobs/5770580210) whenever anything under utils/makemhr or the workflow is updated.
It also uploads artifacts and [(pre)releases it with its own tag](https://github.com/ThreeDeeJay/openal-soft/releases/tag/makemhr) for a [permanent, up-to-date download link](https://github.com/ThreeDeeJay/openal-soft/releases/download/makemhr/makemhr.zip).

### DIFF
--- a/.github/workflows/makemhr.yml
+++ b/.github/workflows/makemhr.yml
@@ -1,0 +1,76 @@
+name: makemhr
+
+on:
+  push:
+    paths:
+      - 'utils/makemhr/**'
+      - '.github/workflows/makemhr.yml'
+
+  workflow_dispatch:
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  Win64:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Get current date
+      run: echo "CurrentDate=$(date +'%Y-%m-%d')" >> $env:GITHUB_ENV
+
+    - name: Get commit hash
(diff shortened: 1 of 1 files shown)


.github/workflows/makemhr.yml
```diff
@@ -0,0 +1,76 @@
...
+    - uses: actions/checkout@v3
+
+    - name: Get current date
+      run: echo "CurrentDate=$(date +'%Y-%m-%d')" >> $env:GITHUB_ENV
+
+    - name: Get commit hash
+      run: echo "CommitHash=$(git rev-parse --short=7 HEAD)" >> $env:GITHUB_ENV
+
+    - name: Clone libmysofa
+      run: git clone https://github.com/hoene/libmysofa.git libmysofa
```
kcat, line 28: It may be a good idea to add `--depth 1` here since it doesn't need to get the entire history. It may also be a good idea to grab a specific release commit so that it's not prone to build failures or run-time bugs that may creep into the latest commit.

ThreeDeeJay: Makes sense
Workflow updated to grab only the latest release tag https://github.com/hoene/libmysofa/releases/tag/v1.3.1

PR #785, opened 2022-11-14
Force find_package to use FindPackage file provided by openal

FranzPoize:
There is an issue if a cmake config file is found and CMAKE_FIND_PACKAGE_PREFER_CONFIG is ON. Cmake will use the config file instead of the findXXX file provided by openal which may not provide the proper cmake variables.

See https://github.com/conan-io/conan-center-index/issues/13772 and https://github.com/conan-io/conan/issues/12488

This is insure that the find_package use the findXXX file provided by openal.

Another idea would be to do the same things that is done with oboe and test if a config file was found (by checking for the proper target) and then default back to MODULE. @kcat if you'd prefer that I can look into it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -880,7 +880,7 @@ set(ALC_OBJS  ${ALC_OBJS}
 
 # Check ALSA backend
 option(ALSOFT_REQUIRE_ALSA "Require ALSA backend" OFF)
-find_package(ALSA)
+find_package(ALSA MODULE)
 if(ALSA_FOUND)
     option(ALSOFT_BACKEND_ALSA "Enable ALSA backend" ON)
     if(ALSOFT_BACKEND_ALSA)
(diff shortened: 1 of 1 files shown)


kcat:
I've always considered OpenAL Soft's Find*.cmake modules to be either stop-gaps or fallbacks for when the library in question doesn't provide one, or isn't otherwise available. If the external project provides its own, OpenAL Soft's should be made compatible with it (if it can't just be removed, depending on how long the external project has had it).

kcat:
Basically I mean like I did with commit 49be7c53cc8d7b8ff15d9fad8349f85b3c96d804. To make `FindPulseAudio.cmake` to set and expose the same variables as `PulseAudioConfig.cmake` (as far as OpenAL Soft uses them, at least). So it doesn't matter whether it uses OpenAL Soft's internal find module, or PulseAudio's own config module, `CMakeLists.txt` will find PulseAudio and use it all the same.

PR #800, opened 2022-12-26
clang-tidy cleanups

neheb:


alc/alu.cpp
```diff
@@ -866,7 +866,7 @@ void CalcPanningAndFilters(Voice *voice, const float xpos, const float ypos, con
          * scaling.
          */
         std::transform(coeffs.begin(), coeffs.end(), coeffs.begin(),
-            std::bind(std::multiplies<float>{}, _1, (1.0f-coverage)*scales[0]));
+            [coverage, scales](float a){ return a * ((1.0f-coverage)*scales[0]); });
```
kcat, line 869: `scales` is an `const std::array&`, so that should be taken by reference.

neheb: good point.

al/auxeffectslot.cpp
```diff
@@ -734,7 +734,7 @@ START_API_FUNC
     switch(param)
     {
     case AL_EFFECTSLOT_GAIN:
-        if(!(value >= 0.0f && value <= 1.0f))
+        if(value < 0.0f || value > 1.0f)
```
kcat, line 737: It's better as it originally was. If `value` is NaN, the original version will catch it, while the new version will pass it (since `nan < 0` and `nan > 1` are both false, it won't be detected as out of range).

neheb: This is an automated change. Sounds like std::isnan should be used.

alc/alc.cpp
```diff
@@ -2630,7 +2630,7 @@ START_API_FUNC
             ProbeAllDevicesList();
 
         /* Copy first entry as default. */
-        alcDefaultAllDevicesSpecifier = alcAllDevicesList.c_str();
+        alcDefaultAllDevicesSpecifier = alcAllDevicesList;
```
kcat, line 2633: It's not pointless. `alcAllDevicesList` is a null-separated list of names, so the string contains names separated by null characters. A copy assignment will copy the full string, including all the names, while using c_str() invokes the `const char*` assignment, which stops at the first null char and copies just the first name.

neheb: Removed.

al/auxeffectslot.cpp
kcat: It's safer to check if the value's not within the valid range/set of values, than to check if it's in the invalid range/set, since the invalid set is unbound. As evidenced by the bug introduced by this change, if a numeric standard adds extant values (e.g. IEEE-754 adding NaN), extra checks need to be added for each that may occur. `std::isnan` makes it three checks instead of two, and who knows what else can show up in the future.

neheb: OK removed.

neheb:
added two commits

alc/alconfig.cpp
```diff
@@ -272,7 +272,7 @@ void LoadConfigFromFile(std::istream &f)
                 ConfOpts.erase(ent);
         }
         else if(!value.empty())
-            ConfOpts.emplace_back(ConfigEntry{std::move(fullKey), expdup(value.c_str())});
+            ConfOpts.emplace_back(std::move(fullKey), expdup(value.c_str()));
```
kcat, line 275: Unfortunately this doesn't work if the struct doesn't have a constructor that takes the given arguments.

core/context.cpp
```diff
@@ -87,7 +87,7 @@ void ContextBase::allocVoiceChanges()
         cluster[i-1].mNext.store(std::addressof(cluster[i]), std::memory_order_relaxed);
     cluster[clustersize-1].mNext.store(mVoiceChangeTail, std::memory_order_relaxed);
 
-    mVoiceChangeClusters.emplace_back(std::move(cluster));
+    mVoiceChangeClusters.push_back(std::move(cluster));
```
kcat, line 90: Is there a reason for this change? `push_back` causes a new object to be default-constructed first and then the given parameter is assigned to it, which is less efficient than `emplace_back` which constructs the new object with the given parameter.

neheb: push_back with std::move is the same really.

kcat: In this particular case after the optimizer takes a stab at it, but not conceptually or in the general case. `push_back` first allocates space for and default-constructs a new object before anything else, then assigns the given parameter to the constructed object (default construction + move assignment). `emplace_back` allocates space for the object, and then constructs the new object using the parameter (move construction). As objects get more complicated, any guarantee of the compiler optimizing `push_back` to act like `emplace_back` goes down, so it's generally better to use `emplace_back` rather than hoping the compiler manages to optimize `push_back` as good as `emplace_back`.

neheb: removed commit.

alc/alconfig.cpp
neheb: removed.

kcat:
If you think it's ready, feel free to click "Ready for review" and I can merge it, unless there's more you still want to do.

SeanTolstoyevski:
hi @neheb 

can you share how to apply with clang-tidy for this project?
any scripts, any configurations, etc.

neheb:
sure. Run

```
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
run-clang-tidy -header-filter=../ -checks=\* > err.txt
```

and filter out the useful fixes that it has. Mostly modernize-, performance-, readability-.

neheb:
@kcat rebased.

PR #809, opened 2023-01-15
constexpr implementation for `pshifter.HannWindow`

SeanTolstoyevski:


kcat:
This doesn't work with C++14, unfortunately, since `std::array`'s `operator[]` and `std::sin` aren't constexpr.

SeanTolstoyevski:
my compiler (gcc 12.2) does not produce any errors.
why?

SeanTolstoyevski:
ao! i understand.
i forgot adding constexpr when create variable :) .

PR #821, opened 2023-02-18
Add explicit disable opensl backend support

halx99:
1. when linking c++_shared, android sdk(23+ tested) will package libOpenSLES.so from ndk into apk, that will cause some old device `slCreateEngine` failed with unkonwn error, i.e. android 5.1.0
2. this change give users of `openal-soft`  opportunity to explicit use single `oboe` backend to avoid above issue


kcat:
The option is already there, by setting `ALSOFT_BACKEND_OPENSL=FALSE`.

halx99:
Thanks.

PR #823, opened 2023-02-21
cleanup switch statements with no case label

maxbachmann:
This fixes warning C4065 on msvc


maxbachmann:
This appears to be done pretty widely. Is this a general code style used in openal, which should stay this way?

kcat:
> This appears to be done pretty widely. Is this a general code style used in openal which should stay this way?

For these instances, yes, because if a public attribute is added for that object and property type in the future, you just need to add a case label for that property to handle it, instead of re-adding the switch statement and putting the error call back in the default case like it already is.

maxbachmann:
:+1: then I will simply suppress the warning in my builds

PR #843, opened 2023-04-25
Decode more FourCC for error searchability

Artoria2e5:
Well. Calling `.c_str()` looks clumsy, but it's either that or a static array like in the SO answer we were looking at. Can I do that? Do we run on multiple threads?

### DIFF
--- a/alc/backends/coreaudio.cpp
+++ b/alc/backends/coreaudio.cpp
@@ -55,6 +55,22 @@ namespace {
 constexpr auto OutputElement = 0;
 constexpr auto InputElement = 1;
 
+std::string InterpretFourCC(UInt32 code)
+{
+    std::string interpreted;
+    for(int i = 3; i > -1; i--) {
+        char ch = (code >> (i * 8)) & 0xff;
+        if(ch > 0x1f && ch < 0x7f)
+            interpreted.push_back(ch);
+        else
+        {
+            interpreted.clear();
+            break;
+        }
+    }
+    return interpreted;
+}
+
 #if CAN_ENUMERATE
 struct DeviceEntry {
     AudioDeviceID mId;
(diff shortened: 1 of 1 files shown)


alc/backends/coreaudio.cpp
```diff
@@ -55,6 +55,22 @@ namespace {
 constexpr auto OutputElement = 0;
 constexpr auto InputElement = 1;
 
+std::string InterpretFourCC(UInt32 code)
+{
+    std::string interpreted;
```
kcat, line 60: Since a UInt32 is a fixed size, you can avoid dynamically generating a `std::string` temporary and use a fixed size array. The kind of thing I do elsewhere would have it like this:
```c++
struct FourCCPrinter {
    char mString[sizeof(UInt32) + 1]{};

    FourCCPrinter(UInt32 code) noexcept
    {
        for(size_t i{0};i < sizeof(Uint32);++i)
        {
            char ch{static_cast<char>(code & 0xff)};
            // If this breaks early it'll leave the first byte null, to get read as a 0-length string.
            if(ch <= 0x1f || ch >= 0x7f)
                break;
            mString[sizeof(UInt32)-1-i] = ch;
            code >>= 8;
        }
    }

    const char *c_str() const noexcept { return mString; }
};
```
Then use it in a similar way:
```c++
ERR("kAudioDevicePropertyStreamConfiguration size query failed: '%s' (%u)\n", FourCCPrinter{err}.c_str(), err);
```
I'd put single-quote marks `'` around the FourCC interpretation, to help distinguish whether it failed to parse and is blank, or it's all spaces or begins/ends with spaces.

kcat:
Superseded by commits e3d5c9f69cba4db2b5d4c0512a1132a83d3ce5d4 and ac143e1236319ea087388ff48a178930c0dbaf47.

PR #848, opened 2023-05-25
Improve oboe backend

halx99:
This change should handle handphone call event properly and fix crash at stop (because the mStream was disconnected by system when handphone call income)

### DIFF
--- a/alc/backends/oboe.cpp
+++ b/alc/backends/oboe.cpp
@@ -5,7 +5,6 @@
 
 #include <cassert>
 #include <cstring>
-#include <mutex>
 #include <stdint.h>
 
 #include "alnumeric.h"
@@ -29,10 +28,17 @@ struct OboePlayback final : public BackendBase, public oboe::AudioStreamCallback
     oboe::DataCallbackResult onAudioReady(oboe::AudioStream *oboeStream, void *audioData,
         int32_t numFrames) override;
 
+    void onErrorAfterClose(oboe::AudioStream* /* audioStream */, oboe::Result /* error */) override;
+
     void open(const char *name) override;
     bool reset() override;
     void start() override;
     void stop() override;
+
+    void restart() {
+        reset();
+        start();
+    }
 };
 
 
@@ -47,6 +53,16 @@ oboe::DataCallbackResult OboePlayback::onAudioReady(oboe::AudioStream *oboeStrea
     return oboe::DataCallbackResult::Continue;
 }
 
+void OboePlayback::onErrorAfterClose(oboe::AudioStream* audioStream, oboe::Result error)
+{
+    // Restart the stream if the error is a disconnect, otherwise do nothing and log the error
+    // reason.
+    if (error == oboe::Result::ErrorDisconnected) {
+        TRACE("Restarting AudioStream");
+        this->restart();
+    }
+    TRACE("Error was %s", oboe::convertToText(error));
+}
 
 void OboePlayback::open(const char *name)
 {
--- a/alc/backends/oboe.cpp
+++ b/alc/backends/oboe.cpp
@@ -34,11 +34,6 @@ struct OboePlayback final : public BackendBase, public oboe::AudioStreamCallback
     bool reset() override;
     void start() override;
     void stop() override;
-
-    void restart() {
-        reset();
-        start();
-    }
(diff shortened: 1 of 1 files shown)


halx99:
refer to: https://github.com/axmolengine/axmol/issues/1212

kcat:
Unfortunately it's not safe to call the backend's `reset` and `start` methods asynchronously to the app like that, as it can change device properties the app can expect not to change. Does the app have a chance to detect a call event from the system and stop OpenAL playback itself before it becomes a disconnection, and resume playback when the call event is done? If not, the proper thing to do may be to call `mDevice->handleDisconnect(...)` to set the AL device as disconnected, where the app can then try to reset it or open a different device if possible.

halx99:
The app no chance to detect call event from system, it's invoke by low-level API of  `AAudio`, refer to:

- https://github.com/google/oboe/blob/main/src/aaudio/AAudioLoader.cpp#L109
- https://github.com/google/oboe/blob/main/src/aaudio/AudioStreamAAudio.cpp#L344

halx99:
About `mDevice->handleDisconnect`, then alsoft user needs set disconnect event callback to handle it, I don't known how to do currently

halx99:
I think I know how to do, will try this API `alcResetDeviceSOFT` later, and I also modify the patch code as your suggestion.

kcat:
There's two ways to detect disconnection: with the `ALC_EXT_disconnect` extension, query `ALC_DISCONNECT`:
```c
ALCint connected;
alcGetIntegerv(device, ALC_CONNECTED, 1, &connected);
if(!connected)
    ... device has been disconnected ...
```
Or using `AL_SOFT_events`:
```c
void MyCallback(ALenum eventType, ALuint object, ALuint param, ALsizei length, const ALchar *message, void *userParam) /* noexcept if C++ */
{
    if(eventType == AL_EVENT_TYPE_DISCONNECTED_SOFT)
        ... device disconnected ...
}
...
// Enable receiving disconnection events
ALenum event = AL_EVENT_TYPE_DISCONNECTED_SOFT;
alEventControlSOFT(1, &event, AL_TRUE);
// Set callback
alEventCallbackSOFT(MyCallback, NULL);
```
Note that for the `AL_SOFT_events` method, you need to dynamically load `alEventControlSOFT` and `alEventCallbackSOFT` using `alGetProcAddress` to use them. Also the callback itself cannot be used to reset the device (you either have to flag it and reset it elsewhere, or send a message to your own event handler to do it elsewhere; you can deadlock if you try to make AL calls in the callback).

kcat (approved):

halx99:
I have try both `alcResetDeviceSOFT` and `alcReopenDeviceSOFT` to ensure the device re-connect, but both of them will crash at `playback->stop`.

halx99:
I have modify oboe backend as follow, then app side use: `alcResetDeviceSOFT`, `alcPauseDevice`, `alcResumeDevice` works, but the audio can't be resume play from last position (interrupted by handphone call event) which opensl backend support.

## oboe.cpp
```cpp
void OboePlayback::onErrorAfterClose(oboe::AudioStream* audioStream, oboe::Result error)
{
    TRACE("Error was %s", oboe::convertToText(error));
    if (error == oboe::Result::ErrorDisconnected) {
        TRACE("Restarting AudioStream");
        mDevice->Flags.reset(DeviceRunning);
        mDevice->handleDisconnect("AudioStream disconnected: %s", oboe::convertToText<oboe::Result>(error));
    }
}
```

## app side
```cpp
void AudioEngineImpl::update(float /*dt*/)
{
   // the _deviceDirty was set to `true` at alsoft device disconnect callback
    if (_deviceDirty.exchange(false, std::memory_order_relaxed)) {
        alcResetDeviceSOFT(s_ALDevice, nullptr);
        //auto proc = (decltype(alcReopenDeviceSOFT)*) alcGetProcAddress(s_ALDevice, "alcReopenDeviceSOFT");
        //proc(s_ALDevice, nullptr, nullptr);
        alcDevicePauseSOFT(s_ALDevice);
        alcDeviceResumeSOFT(s_ALDevice);
    }
    std::unique_lock<std::recursive_mutex> lck(_threadMutex);
    _updatePlayers(false);
}
```

PR #853, opened 2023-05-29
Wasapi uwp support

halx99:


alc/backends/wasapi.cpp
```diff
@@ -1071,11 +1076,24 @@ HRESULT WasapiPlayback::openProxy(const char *name)
         mDevice->DeviceName = std::string{DevNameHead} + name;
     else
         mDevice->DeviceName = DevNameHead + DeviceHelper::get_device_name_and_guid(mMMDev).first;
+
+#if defined(ALSOFT_UWP)
+    mDefaultChangeHandler = MediaDevice::DefaultAudioRenderDeviceChanged +=
+        ref new TypedEventHandler<Platform::Object ^, DefaultAudioRenderDeviceChangedEventArgs ^>(
+            [this](Platform::Object ^ sender, DefaultAudioRenderDeviceChangedEventArgs ^ args) {
+        if (args->Role == AudioDeviceRole::Default)
+            mDevice->handleDisconnect("Default audio device changed %s", wstr_to_utf8(args->Id->Data()).c_str());
```
kcat: I don't think the device should be set to disconnect just because the default changed. As long as the current device is still usable, it should keep working. There app could have a reason for using the device it does and not want to change if the default does, after all.

I do intend to add an API to notify the app when the default device changes (along with possibly device add/remove notifications), which would allow the app to more smoothly change outputs as it wants.

halx99: > I don't think the device should be set to disconnect just because the default changed. As long as the current device is still usable, it should keep working. There app could have a reason for using the device it does and not want to change if the default does, after all.
> 
> I do intend to add an API to notify the app when the default device changes (along with possibly device add/remove notifications), which would allow the app to more smoothly change outputs as it wants.

I have add new event: `AL_EVENT_TYPE_DEFAULT_DEVICE_CHANGED_SOFT`, please help review

alc/alu.cpp
```diff
@@ -2205,3 +2205,29 @@ void DeviceBase::handleDisconnect(const char *msg, ...)
...
+        auto len              = std::min(deviceId.length(), sizeof(default_changed.mDeviceId) - 1);
+        if (len > 0)
+            memcpy(default_changed.mDeviceId, deviceId.data(), len);
+        default_changed.mDeviceId[len] = '\0';
+        default_changed.mDataFlow      = dataFlow;
+
+        for (ContextBase* ctx : *mContexts.load())
+        {
+            RingBuffer* ring{ctx->mAsyncEvents.get()};
+            auto evt_data = ring->getWriteVector().first;
```
kcat, line 2224: Unfortunately it's not that simple. The ringbuffer is single-producer/single-consumer and is primarily used by the mixer to signal source events as they happen, so you can't write events to the ringbuffer outside of the mixer thread while it's running. I've been trying to find a lockless multi-producer/single-consumer queue, but the things I keep running into sneak in a spinlock on the producer side, making it not lockless and not suitable for the mixer to use.

halx99: It's need your help, you can write this pr branch of my fork directly.

kcat: It may be best to not bother with default device change detection for now, since I'm not sure how to do it yet. If I can't get a real-time-safe multi-producer queue working, I may need to do some kind of event marshaling to the mixer which can then put it in the event queue (which isn't straightforward, since the mixer may not be running).

Alternatively, I may set up a separate ALC event API for system-level events, which doesn't have to contend with AL source events being generated in real-time (and could also better handle race conditions, e.g. the app opening the default device, the default device changes, then the app sets up the callback to get event notifications, leaving the app unaware that the default device changed).

halx99: Maybe add a flag(indicate whether default device changed) to backend and check it in  mixer thread and fire the event?

kcat: That's one possibility. Though it would need to handle the case of the mixer not running (either because there's no context, the device has been paused, or is disconnected), which can start or stop asynchronously to receiving the system event.

Another possibility is to use the debug API, which isn't as restricted with adding entries to it. It's not as much intended to indicate specific system events, but it's not out of the question.

But I'm thinking a separate ALC API would be better for handling system-generated events, leaving AL_SOFT_events for events generated from mixing. Being in the ALC API would also allow tracking the events without an active device or context, avoiding a race condition.

halx99: > But I'm thinking a separate ALC API would be better for handling system-generated events, leaving AL_SOFT_events for events generated from mixing. Being in the ALC API would also allow tracking the events without an active device or context, avoiding a race condition.

OK, I have done my changes on branch [wasapi-uwp-support](https://github.com/halx99/openal-soft/tree/wasapi-uwp-support)

halx99:
Nice, I will try to use your system events

alc/alu.cpp
halx99: I have done notify default device change by `alc::Event`

alc/backends/wasapi.cpp
```diff
@@ -190,159 +199,505 @@ bool checkName(const al::span<DevMap> list, const std::string &name)
...
+#if defined(ALSOFT_UWP)
+    /** ----------------------- IActivateAudioInterfaceCompletionHandler ------------ */
+    HRESULT ActivateCompleted(IActivateAudioInterfaceAsyncOperation* operation) override
+    {
+        HRESULT hrActivateResult     = S_OK;
+        IUnknown* punkAudioInterface = nullptr;
 
+        HRESULT hr = operation->GetActivateResult(&hrActivateResult, &punkAudioInterface);
+        // Check for a successful activation result
+        hr = operation->GetActivateResult(&hrActivateResult, &punkAudioInterface);
```
kcat, line 332: Did you mean to call this twice?

halx99: Fixed

halx99:
Maybe the related issues can be close as complete.

PR #466, opened 2020-08-15
macOS osx/ios dynamic framework support

halx99:

PR #854, opened 2023-05-31
Add ios ci check

halx99:
@kcat  The latest code has compile error for ios build,

```txt
/Users/runner/work/openal-soft/openal-soft/alc/backends/coreaudio.cpp:44:10: fatal error: 'IOKit/audio/IOAudioTypes.h' file not found
```

refer to: https://github.com/halx99/openal-soft/actions/runs/5133473000/jobs/9236097000#step:5:385

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,18 @@ jobs:
             build_type: "Release"
           }
         - {
+            name: "iOS-Release",
+            os: macos-latest,
+            cmake_opts: "-GXcode \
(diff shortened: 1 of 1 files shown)


halx99:
> Is that how you build for iOS, just set `-DCMAKE_SYSTEM_NAME=iOS` and `-DCMAKE_OSX_ARCHITECTURES=arm64` for CMake? I'd have expected it to be a bit more involved, with needing an iOS-specific toolchain environment (different headers, libs, etc, and other CMake properties that need changing).

Yes, refer to: #466

PR #855, opened 2023-06-01
Fix ios crash at sem_init fail with errno=78(function not implemented)

halx99:


common/threads.h
```diff
@@ -16,8 +16,10 @@
 
 #if defined(__APPLE__)
 #include <AvailabilityMacros.h>
-#if (MAC_OS_X_VERSION_MIN_REQUIRED > 1050) && !defined(__ppc__)
+#include <TargetConditionals.h>
```
barracuda156, line 19: I am not sure `TargetConditionals.h` exists in every MacOS SDK. This has to be verified, otherwise it will break the build.

barracuda156:
@halx99 Allow me to test it, please (will do in an hour or so, once GCC build is complete and machine is free), but otherwise looks good.

common/threads.h
halx99: Yes, the `TargetConditionals.h` exists on all modern apple platforms

barracuda156:
@halx99 Yes, your fix works fine on 10.6 PPC. Should be good to merge.

@kcat Turned out, inttypes fix in CMakeLists does not work: https://github.com/kcat/openal-soft/pull/851#issuecomment-1571810525
Sorry, I just assumed that time it will do what it is supposed to without rebuilding with it -_-

UPD. I know why it fails! You have used `cinttypes` header in the test, but source code uses `inttypes.h`. The first header does not need `__STDC_FORMAT_MACROS` with GCC, the second does. We should either change the test or the source.

halx99:
@kcat This pr also ready to be merge.

sgowen:
FWIW, I've just verified that the crasher on iOS is fixed if I apply the changes from this PR

barracuda156:
@kcat Would you fix the header in the inttypes check or should I make a PR for that?

kcat:
I think I'd prefer having the sources use `cinttypes` instead. I can do it.

barracuda156:
@kcat Great, thank you!

PR #859, opened 2023-06-08
Fix linker error with undefined symbol '=' on MSVC

Rynnya:
Reason for such symbol to ever exist is `operator==` for GUID/IID (see guiddef.h)
CMake exports such symbol into exports.def where LINK reads it and (obviously) does not found it, which cause linker error
Such error only happens with Debug build, because `operator==` and `operator!=` isn't inlined
Replacing every occurrence of `operator==` (AND `operator!=` because it uses `operator==` internally) with `IsEqualGUID` or `IsEqualIID` fixes this issue

For some unknown to me reason using `operator==` in `AL` is fine and doesn't cause such behaviour to happen (i believe this happens because it's exported manually instead of automatic CMake export)

### DIFF
--- a/alc/backends/dsound.cpp
+++ b/alc/backends/dsound.cpp
@@ -331,7 +331,7 @@ void DSoundPlayback::open(const char *name)
             hr = CLSIDFromString(utf8_to_wstr(name).c_str(), &id);
             if(SUCCEEDED(hr))
                 iter = std::find_if(PlaybackDevices.cbegin(), PlaybackDevices.cend(),
-                    [&id](const DevMap &entry) -> bool { return entry.guid == id; });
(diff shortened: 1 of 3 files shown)


kcat:
What version of MSVC does this happen with? Can you paste the error output?

When a function marked inline (implicitly or explicitly) isn't inlined, it's supposed to emit a callable copy of the function, so it should still link. It seems like something else is going on if there's a failure here.

Rynnya:
MSVC version is 19.34.31937 (Visual Studio 2022, 17.4.4)

Building OpenAL in stand-alone mode doesn't cause such issue, but building it as subdirectory for some reason fails every time
image

Running `dumpbin /symbols dsound.obj` shown following exports
```
922 00000000 SECT40B notype ()    External    | IsEqualGUID
923 00000000 SECT7  notype ()    External     | ==
```

Issue #857, opened 2023-06-02
[bug] windows unhandled exception


hsdk123:
Hi, I need to do some debugging, but it seems like a recent commit here has started causing a crash on windows:

![image](https://github.com/kcat/openal-soft/assets/4100142/2affc693-8af3-4cd3-92a8-7e067f12163e)

Confirming things work until `6cd3d2414ac40a94713c84a0703d3e021bf9a570`

kcat:
Commit 6cd3d2414ac40a94713c84a0703d3e021bf9a570 shouldn't have any bearing on Windows/WASAPI behavior. It shouldn't be functionally different at all, and neither the previous or following commit. Are you able to get a backtrace?

hsdk123:
@kcat I think you've misread my comment, I was stating that things [work] until Commit https://github.com/kcat/openal-soft/commit/6cd3d2414ac40a94713c84a0703d3e021bf9a570. Not sure what commit after creates the bug. 

I'll try get a backtrace

Rynnya:
Reventing commit https://github.com/kcat/openal-soft/commit/1a583b016a7b5e22f370e2f425722e3dc240a7dd fixes this issue

kcat:
I really don't like COM. Try with commit d769041916df4efb48fd212683dc9842decc10f1.

Rynnya:
Issue gone after applying https://github.com/kcat/openal-soft/commit/d769041916df4efb48fd212683dc9842decc10f1, program didn't crash anymore, thank you!

hsdk123:
Confirmed as well, thanks! Would be great to add a test around this if possible

kcat:
There is `openal-info` and the other utilities/examples. I'm not sure how to do automated tests as part of CI, though.

PR #860, opened 2023-06-11
Add gtest integration

hsdk123:
Hi, this is a PR for adding gtest integration, in reply to 

https://github.com/kcat/openal-soft/issues/857

https://github.com/hsdk123/openal-soft/actions/runs/5237440993/jobs/9455621772

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,7 @@ jobs:
             cmake_opts: "-A x64 \
               -DCMAKE_SYSTEM_NAME=WindowsStore \
               \"-DCMAKE_SYSTEM_VERSION=10.0\" \
+              -DALSOFT_TESTS=OFF \
               -DALSOFT_BUILD_ROUTER=ON \
               -DALSOFT_REQUIRE_WASAPI=ON",
             build_type: "Release"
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
```
hsdk123, line 3: `pull_request` is what has the CI checks run + show on this PR

.gitignore
```diff
@@ -2,6 +2,8 @@ build*/
 winbuild
 win64build
 .vs/
+CMakeCache.txt
+CMakeFiles/
```
kcat, line 6: These two `CMake*` entries don't seem necessary. It shouldn't be built in the source tree like that, and there's a number of other files it would create there if it was.

CMakeLists.txt
```diff
@@ -111,6 +111,7 @@ option(ALSOFT_UTILS "Build utility programs"  ON)
 option(ALSOFT_NO_CONFIG_UTIL "Disable building the alsoft-config utility" OFF)
 
 option(ALSOFT_EXAMPLES  "Build example programs"  ON)
+option(ALSOFT_TESTS "Build test programs"  ON)
```
kcat, line 115: I don't think this should be on by default, since it tries to download extra stuff and I don't think most builds would need or want it.

hsdk123: updated!

.gitignore
hsdk123: updated!

PR #868, opened 2023-06-28
[UWP] Fix crash when probe capture device fail

halx99:


alc/backends/wasapi.cpp
```diff
@@ -674,6 +674,8 @@ struct DeviceHelper final : private IMMNotificationClient
         const auto deviceRole = Windows::Media::Devices::AudioDeviceRole::Default;
         auto DefaultAudioId   = flowdir == eRender ? MediaDevice::GetDefaultAudioRenderId(deviceRole)
                                                    : MediaDevice::GetDefaultAudioCaptureId(deviceRole);
+        if (!DefaultAudioId)
+            return defaultId;
```
kcat, line 678: It may be a good idea to still do the enumeration even if it can't get the default. Also, how does this behave with `openDevice`? If it also gets a nullptr for the default ID, it calls `DeviceInformation::CreateFromIdAsync` too.

halx99: Also crash at `DeviceInformation::CreateFromIdAsync`

halx99: And usually, the audio render device not `null`

PR #887, opened 2023-07-26
disable unavailable functionality on xbox

maxbachmann:
This disables functionality that is not available on the xbox similar to `ALSOFT_UWP`.

### DIFF
--- a/common/strutils.cpp
+++ b/common/strutils.cpp
@@ -45,7 +45,11 @@ namespace al {
 
 std::optional<std::string> getenv(const char *envname)
 {
+#ifdef _GAMING_XBOX
+    const char *str{::getenv(envname)};
+#else
     const char *str{std::getenv(envname)};
+#endif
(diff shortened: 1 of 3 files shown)


common/strutils.cpp
```diff
@@ -45,7 +45,11 @@ namespace al {
 
 std::optional<std::string> getenv(const char *envname)
 {
+#ifdef _GAMING_XBOX
+    const char *str{::getenv(envname)};
+#else
     const char *str{std::getenv(envname)};
+#endif
```
maxbachmann, line 52: For unknown reason `std::getenv` is not available for non desktop apps:
```
#ifdef _CRT_USE_WINAPI_FAMILY_DESKTOP_APP
using _CSTD getenv;
using _CSTD system;
#endif // _CRT_USE_WINAPI_FAMILY_DESKTOP_APP
```
however `getenv` is available

PR #919, opened 2023-09-25
Make cppwinrt version configurable

halx99:


CMakeLists.txt
```diff
@@ -1409,14 +1409,16 @@ else()
     target_link_libraries(${IMPL_TARGET} PRIVATE alcommon ${LINKER_FLAGS} ${EXTRA_LIBS} ${MATH_LIB})
 
     if(ALSOFT_UWP)
+        set(ALSOFT_CPPWINRT_VERSION "2.0.230706.1" CACHE STRING "The soft-oal default cppwinrt version")
```
kcat, line 1412: Is this possible to do without caching the version? If the version is updated in the future, existing builds will keep the old version in cache even if they never set it.

halx99: The `CACHE STRING` behavior same with option for `bool` type, user can pass `cmake "-DALSOFT_CPPWINRT_VERSION+xxx" to overwrite it, so i think it's ok.

PR #920, opened 2023-09-25
Compile with c++20 support

halx99:
compiler errors without this patch, refer to: https://github.com/kcat/openal-soft/actions/runs/6275514636/job/17043155913

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ endif()
 
 set(DEFAULT_TARGET_PROPS
     # Require C++17.
-    CXX_STANDARD 20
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ endif()
 
 set(DEFAULT_TARGET_PROPS
     # Require C++17.
-    CXX_STANDARD 17
+    CXX_STANDARD 20
```
kcat, line 80: I don't think I like this, since this could allow using C++20 features accidentally when C++20 isn't required yet.

Issue #862, opened 2023-06-21
Feature request: setting ALSOFT_LOGFILE by app


Lulu04:
Hi, thanks for sharing OpenAL Soft. Sorry for my bad english.

At the moment, to set the path and filename for the log file, we have to declare an environment variable or changing a line in the config file.
This action must be done by the user and it is not convenient for an app that use OpenAL-Soft to ask its user to sets this variable.

There is also the problem when two application use openAL Soft:
  - the user start the first one, a log file is created by the AL
  - the user quits the first application and starts the second: the first log file is erased and the debug info are lost.
 
In addition, setting this variable programatically seem to be not easy under Linux (however I'm not a Linux expert).

It would be usefull if the app itself can set this path+filename for the log file when it start. This way, both the app log file and OpenAL log file can be to the same directory, easy to retrieve for debug purpose.

Regards

kcat:
Sorry for the delayed response.

For the time being, I've added a function to set a callback that gets the log messages:
```c++
typedef void (ALC_APIENTRY*LPALSOFTLOGCALLBACK)(void *userptr, char level, const char *message, int length) noexcept;
void ALC_APIENTRY alsoft_set_log_callback(LPALSOFTLOGCALLBACK callback, void *userptr) noexcept;
```

This function isn't directly exported by the library, so you have to load it with `alcGetProcAddress(NULL, "alsoft_set_log_callback");`, then call it with a non-null callback before calling any other OpenAL function. The callback will get a `level` parameter which is a character indicating the severity of the message (currently `'I'` for informational, `'W'` for non-severe warning, and `'E'` for error; more may be used in the future), a 'message' pointer to the null-terminated string, and the `length` of the string (excluding the null char).

The callback may be called on various different threads. It uses a mutex internally to prevent being called from multiple threads at the same exact time, but different calls can be on different threads.

If a callback isn't set before messages start getting logged, the callback won't ever be called if set later. You also must not call OpenAL functions in the callback.

Lulu04:
Thanks you very much for implementing this feature.
I've just tested it and it works like a charm!
[GiveYourVoice.log](https://github.com/kcat/openal-soft/files/12016914/GiveYourVoice.log)

Thanks again for your great work
Regards

PR #939, opened 2023-11-27
Move logging callback into an OpenAL extension

MathiusD:
The aim here is to enable access to a logging callback through an OpenAL extension (here ALC_SOFT_logging_events).

Logging callback has already been implemented here: #862.

The only difference with the previously implemented callback is the following modifications:

The log level is now represented by an ALCenum
Two new parameters are added:
 * A null terminated string containing the name of the entity issuing the log.
 * The size of the previous parameter

This string, representing the emitter, can be null, if the implementation cannot provide an emitter section. The emitter represents the impacted section, aka the name of the extension, name of method, or name of specification from which the log comes.

### DIFF
--- a/alc/alc.cpp
+++ b/alc/alc.cpp
@@ -308,7 +308,8 @@ constexpr ALCchar alcNoDeviceExtList[] =
     "ALC_SOFT_loopback "
     "ALC_SOFT_loopback_bformat "
     "ALC_SOFT_reopen_device "
-    "ALC_SOFT_system_events";
+    "ALC_SOFT_system_events "
+    "ALC_SOFT_logging_events";
 constexpr ALCchar alcExtensionList[] =
     "ALC_ENUMERATE_ALL_EXT "
     "ALC_ENUMERATION_EXT "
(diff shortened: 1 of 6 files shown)


kcat:
I'm not keen on making this an extension since it's dealing with implementation details of what the library's doing, things which can change. The app isn't supposed to gleam any details from it, it's just a way to include OpenAL Soft's logging messages in the app's own log, to avoid the user having to mess with the `ALSOFT_LOGLEVEL`/`ALSOFT_LOGFILE` env var stuff.

There is instead the `AL_EXT_debug` API for the app to get OpenAL-related messages (primarily just API errors for now, with some portability warnings).

MathiusD:
Okay, I understand (and to tell the truth, I hadn't explored `AL_EXT_DEBUG` enough ^^')

However, I'd like to take this opportunity to ask another question: Is it desirable to expose these implementation-dependent methods (i.e. all those in inprogrext.h that don't concern an extension or an ABI) in a public include? (For example, in `include/SOFT/api.h`) (An example of what I mean here : https://github.com/kcat/openal-soft/compare/master...mathiusd:feature/move_non_standard_export_into_dedicated_header) (I ask because it allows anyone to manage includes more finely).

kcat:
Functions that don't have `AL_API` or `ALC_API` aren't exported from the library, so a function declaration in a public header isn't too appropriate. I also prefer to limit the number of exported non-essential functions (non-standard, and what an app won't require to work).

I'd really like a way to export a function in a way that it can be retrieved with `dlsym`/`GetProcAddress`, but not used directly to access at link-time. There doesn't seem to be a way to do that though, and putting it in a public header would make it easy to make the function required at run-time.

MathiusD:
Okay, thanks a lot for the explanation!

PR #940, opened 2023-11-29
feat(openal-info): Add values related to sources inside openal-info.c

MathiusD:
Small PR to add sources values inside openal-info

### DIFF
--- a/utils/openal-info.c
+++ b/utils/openal-info.c
@@ -124,12 +124,6 @@ static ALCenum checkALCErrors(ALCdevice *device, int linenum)
 }
 #define checkALCErrors(x) checkALCErrors((x),__LINE__)
 
-static void printALCIntegerValue(ALCdevice *device, ALCenum enumValue, char* enumName)
-{
-    ALCint value;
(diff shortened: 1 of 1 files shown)


utils/openal-info.c
```diff
@@ -148,6 +154,10 @@ static void printALCInfo(ALCdevice *device)
         printf("ALC extensions:");
         printList(alcGetString(device, ALC_EXTENSIONS), ' ');
         checkALCErrors(device);
+        printALCIntegerValue(device, ALC_MONO_SOURCES, "ALC_MONO_SOURCES");
+        checkALCErrors(device);
+        printALCIntegerValue(device, ALC_STEREO_SOURCES, "ALC_STEREO_SOURCES");
```
kcat, line 159: This should be done after the context is created and made current. It would also make sense to put the `checkALCErrors` call by the `alc*` calls that could cause an error, and avoid printing the values if there was an error since the variable would be uninitialized in that case.

MathiusD: Yes, my bad :x, it's fixed

PR #1002, opened 2024-06-09
Use std::to_address if available

georgthegreat:
Prevents 
```
common/alspan.h:105:57: error: call to 'to_address' is ambiguous
    constexpr explicit span(U iter, index_type) : mData{to_address(iter)} { }
```

when `std::to_address` can be found via ADL.

### DIFF
--- a/common/almalloc.h
+++ b/common/almalloc.h
@@ -88,6 +88,9 @@ constexpr bool operator!=(const allocator<T,N>&, const allocator<U,M>&) noexcept
 { return allocator<T,N>::Alignment != allocator<U,M>::Alignment; }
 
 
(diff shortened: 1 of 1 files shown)


kcat:
I think it would be better to fix the ambiguous call rather than try to sometimes use the standard function, which can have behavioral differences and cause different builds to work differently. Once the project switches to C++20, it'll use `std::to_address` unconditionally instead of the custom one.

georgthegreat:
`std::to_address` is just way to simple to differ in any way.
But if you prefer using your implementation, please, fix it ASAP, as we are unable to upgrade to the current version.

georgthegreat:
Thanks, @kcat!

Issue #757, opened 2022-08-29
Add some (non-default) option to hit an hotpath of the original router device selection


mirh:
See https://github.com/kcat/openal-soft/issues/650#issuecomment-1186326473.
In order to play nice with Creative's official OAL router (and yes, this seems the only loophole to avoid recurrent maintenance of either system or games folders) the device name is key. 
https://github.com/kcat/openal-soft/blob/b8d73a226ab4251f94a9096fa95a20fac58e826b/alc/alc.cpp#L919
The best way to this aim, if not any because it would be entirely addressable within openal-soft, would be for its name to always match whatever the current default device in windows is (as per K971's observations). But AFAICT this would break total mayhem in `ALC_ENUMERATE_ALL_EXT` querying. 
...unless perhaps the same knob also enabled "soundblaster-like behaviour" (i.e. only care about a single device, don't report "XX on YY")?

And so the next (and only other possible) solution would be to rely on the "partial name match" path. 
I'm not 100% sure on how exactly they filter the list of the available implementations depending on the name of the current device.. but anyway long story short, if both the device and the openal driver have "X-Fi" or "Audigy" in the name they get more priority than the godawful "Generic Software" fallback. 
And so, rather than just relying on some other kind of hardcoded strings, I thought that we could as well just use an environmental variable to arbitrarily change the name (so that you wouldn't have to include in the repo those names, or other awfully similar puns). 
Of course the user would also have to be instructed to rename their playback device, but this is another matter. 

p.s. I had originally started this feature request to just propose the second idea, but throughout the writing of the premise (and of the logical implications that would entail there aren't other ways to make OpenAL32.dll happy) the first oddball emerged too. 
I'm not asking for both, or even one of them in particular.. but yeah. For the moment, after all this brainstorming I'd just like to receive a reality check if I haven't missed anything else.

kcat:
Spoofing this "properly" turns out to be a bit more difficult than anticipated. Adding a unique user-specified name to the enumeration lists isn't that hard, nor would it be hard to open the default device when trying to open it. However, if the spoofed name is opened, having that device report back that spoofed name as its name is a bit more involved.

[...]

Issue #682, opened 2022-04-10
[Feature Request] Low latency output on Windows via ASIO/WASAPI exclusive


ThreeDeeJay:
High audio latency (100ms+) is something that has plagued apps/games on Windows for a long time, yet people rarely notice, let alone measure it, like [Matt Gore, HeSuVi developer](https://sourceforge.net/p/hesuvi/wiki/Audio%20Latency%20Measurements/) and [Battle(non)sense](https://youtu.be/Kggh_Ge8YHs?t=352).
![image](https://i.imgur.com/ZKy640S.jpeg)
![image](https://i.imgur.com/aefWeum.jpeg)

So I've been wondering if ASIO could be implemented into OpenAL Soft directly, since [Crystal Mixer already did something like that](https://sourceforge.net/projects/crystal-mixer/files/Crystal%20Mixer/), but AFAIK it's only capable of virtualizing the multichannel audio mix.

Alternatively, someone even modified OpenAL Soft to use WASAPI in exclusive mode, which I've tested and confirmed it does make a difference (tho I'm yet to measure it), so I forked it [here](https://github.com/kcat/openal-soft/compare/master...ThreeDeeJay:openal-soft:WASAPI-Exclusive)
Perhaps a flag to switch to exclusive mode in the main branch would be more feasible since it just seems to require a couple line edits (tho it would probably need some improvement so it's not restricted to `sample-type=int16` and `period_size` is automatically set to the lowest supported by the sound card, as well as minimal/no mixahead or any other bottlenecks to ensure lowest possible latency).

Either option should hopefully allow ultra low latency on [thousands of games that are at least potentially supported by OpenAL Soft](https://airtable.com/shryB7ft0Cle17cLA) by using the sound card's native ASIO or ASIO4ALL. I think audio would also be bit-perfect audio (at least on WASAPI exclusive) by bypassing the Windows mixer. So perhaps eventually including both would give people options based on their needs.

kcat:
I'm curious how much of the latency is a result of using shared mode or non-ASIO output. "Button to audio" latency with random games doesn't say much, since it's also including input latency (the time from physically pressing a button to the OS detecting the input, then to the process detecting the input), and logic/frame latency (the time from the process getting input to processing a new logic frame, and from a logic frame to updating audio state, which can be at different rates), and only then getting to the audio latency. 

[...]

Issue #586, opened 2021-07-20
[Feature request] Spatial renderer for Microsoft Spatial Sound API


ThreeDeeJay:
![image](https://user-images.githubusercontent.com/71472458/126266074-6daecfa3-c5e4-4d58-bffb-ab6ac0cd3cca.png)
OpenAL Soft as a spatial sound option via the Windows 10's built-in API could (in theory) allow:
- Personalized/preferred HRTFs consistent with other games using MHR-based HRTF (via [OpenAL](https://github.com/kcat/openal-soft)/[DirectSound3D](https://github.com/kcat/dsoal)/[X3DAudio](https://github.com/kosumosu/x3daudio1_7_hrtf)). Sonic is lackluster and at least for me, _really_ inaccurate. Atmos and DTS sound better but I don't think they're worth their price tag. [Comparison here](https://kutt.it/BinauralHRTF).
- Automatically enable virtual surround, only for games/programs outputting surround, while preserving stereo content (e.g. music apps) unaltered, which system-wide virtual surround isn't capable of, unless we use a virtual audio device for virtual surround that also allows outputting stereo to the real sound card directly.
- Bypass requirement of a surround-capable sound card or virtual audio device for virtual surround, like when using a Equalizer APO + HeSuVi setup, which can potentially add some audio delay, bugs, hassle to set up for people unfamiliar with it.

kcat:
You mean as some kind of "driver" for the Spatial Audio API, such that any app using that API will implicitly use OpenAL Soft's renderer under the hood? Is there any information about how to make one of those?

ThreeDeeJay:
Exactly. And I'm not sure this is enough information to actually implement it, but it seems relevant:

[Microsoft Spatial Sound and Audio Middleware](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/CoreAudio/spatial-sound.md#microsoft-spatial-sound-and-audio-middleware)
[...]

PR #1008, opened 2024-06-26
Release portable builds

ThreeDeeJay:
Just an update for the GitHub Actions CI workflow to (in addition to artifacts already being uploaded to the workflow page) release Windows 32 and 64-bit artifacts in a portable package (soft_oal.dll renamed to OpenAL32.dll with an optional, separate package pre-configured for HRTF, and documentation) for simple drag and drop local replacement for specific games/apps to make it easier for users who are not familiar with [copying files to system folders](https://www.pcgamingwiki.com/wiki/Glossary:Surround_sound#OpenAL), ensure OpenAL Soft is used instead of other libraries and avoid issues with the router (e.g. currently on my machine, when using the latest router and driver DLLs, the [EFXShow](https://openal.org/downloads/OpenAL11CoreSDK.zip) process hangs before starting even though renaming soft_oal.dll to replace the OpenAL32.dll router works, both in the system folder and locally)

I pushed the HRTF config to configs/HRTF/alsoft.ini, so perhaps in the future we could also have other pre-made configs like say, 7.1.4/Spatial Audio API, surround sound, crossfeed, super stereo, wave writer, etc., but let me know if it'd be better to use a different location.

My only concern is the issues mentioned [here](https://www.pcgamingwiki.com/wiki/Glossary:Surround_sound#OpenAL) when not using the router: 
> DO NOT rename the soft_oal.dll files and use them as a substitute OpenAL32 file when doing steps 4 and 5. The method provided above is bound to run into far less problems by using a method similar to the way in which the original OpenAL executed it's dlls, alongside allowing for newer extensions.

Even though the official, stable release readme is much more lenient:

> If you do not wish to use the router, then rename soft_oal.dll to OpenAL32.dll
(note: even the 64-bit DLL should be named OpenAL32.dll). Just be aware this
will prevent other system-installed OpenAL implementations from working.

So does the router really provide any fixes/improvements compared to using the OpenAL Soft driver library directly?
And if so, would it be possible to load both the router and driver DLL locally, perhaps even forcing OpenAL Soft library?
The router is refusing to work on my end so I can't check.

Anyhow, this should allow downloading the latest builds directly from static, permanent links like:
https://github.com/ThreeDeeJay/openal-soft/releases/tag/latest
https://github.com/ThreeDeeJay/openal-soft/releases/download/latest/OpenALSoft.zip
https://github.com/ThreeDeeJay/openal-soft/releases/download/latest/OpenALSoft+HRTF.zip
image
Note that apparently new tags like [1.23.1](https://github.com/kcat/openal-soft/releases/tag/1.23.1) didn't transfer to [my fork](https://github.com/ThreeDeeJay/openal-soft/releases/tag/1.23.1) when syncing to the main repo, but hopefully it should always [use the latest tag that matches the #.#.# format](https://github.com/ThreeDeeJay/openal-soft/blob/042ad4c3da585a4019f3bdd8fe5cc8e923e226f1/.github/workflows/ci.yml#L118). 🤞👀💦


configs/HRTF/alsoft.ini
```diff
@@ -0,0 +1,74 @@
+[general]
```
kcat, line 1: This should probably only include necessary settings, rather than all possible ones. Especially ones like `resampler`, `output-limiter`, and `dither`, which aren't relevant to HRTF.

ThreeDeeJay: Ohh, I just found out `cubic` (recommended [here](https://youtu.be/62U6UnaUGDE)) was [renamed to `gaussian`](https://github.com/kcat/openal-soft/commit/3ad68aa979de387fa05efad0a081a064c382580b) so it indeed makes sense to exclude it. 👌 

`output-limiter` was set to hopefully address [sounds playing so loud it clips](https://www.youtube.com/watch?v=2ed_mUTmIvs&t=12170s) but even though I haven't noticed that happening after enabling the limiter,  I admit I haven't confirmed whether that really did fix it, or if that that was just a bug with some sounds or DSOAL where a limiter wouldn't help. If there's a better/proper fix, let me know but for now I'll just exclude this option. :

I forgot who/why they recommended `dither` but that one can go too.

And as for the empty flags, I just included them so that the user can easily set values without having to dig up the flags from the template config, but don't they get ignored by the global/default config? 🤔

kcat: > And as for the empty flags, I just included them so that the user can easily set values without having to dig up the flags from the template config, but don't they get ignored/overridden by the global/default config? 🤔

App-local configs override global ones, and setting a blank value clears whatever value may have been already set. Having `resampler=` in an app-local config file, for example, will cause whatever resampler may be set in a global config to be overridden/ignored and use the default.

The available config options also change with time, with new ones being added and old ones being deprecated or removed, which risks that file getting out of sync if it has unnecessary options. They also lack details and explanations about what the options are and what values can be set, so a user may need to look through the template/sample config file to know what to do with it anyway.

ThreeDeeJay: Ah I get it now.
How about [commenting out the empty flags](https://github.com/ThreeDeeJay/openal-soft/commit/8865dcdfcd11d61632c3faf5d7dd206347626981)?
image


That way users can find flags quickly when instructed to set a specific value.
I also made the workflow [add the config template to the Documentation folder](https://github.com/ThreeDeeJay/openal-soft/commit/e7c0736f7e5524ea655ecca0a510ff7a4a31a23f) for those who want to learn what each flag does. 👌 

image
image
image

kcat: > Ah I get it now.
> How about [commenting out the empty flags](https://github.com/ThreeDeeJay/openal-soft/commit/8865dcdfcd11d61632c3faf5d7dd206347626981)?

I'd still prefer the unused settings to be removed, to avoid extra maintenance with new or deprecated options. As it is, I'm not sure the config file is even really needed, since WASAPI shouldn't have an issue detecting when the device is set to headphones, and it will just force HRTF when the device doesn't report being headphones. Forcing HRTF when the device isn't headphones, or even if it is headphones but is using virtual surround sound to simulate surround sound speakers, won't sound good and it's better to keep it as speaker output.

FWIW, the `hrtf` option is unneeded (it's deprecated, `stereo-mode=headphones` and `stereo-encoding=hrtf` replace it), and `hrtf-mode=full`, `hrtf-size=0`, and `cf_level=0` are the defaults.

> That way users can find flags quickly when instructed to set a specific value.
> I also made the workflow [add the config template to the Documentation folder](https://github.com/ThreeDeeJay/openal-soft/commit/e7c0736f7e5524ea655ecca0a510ff7a4a31a23f) for those who want to learn what each flag does. 👌

I don't think having unused ones commented out will help much for users being instructed to add settings, since it's not any harder to say to "add `foo=bar` under `[general]`" compared to "uncomment `foo=` and set it to `foo=bar`". Having the template for people who want to read about possible config options is fine, but isn't really helped by having the options commented out in the actual config file. Especially if the real config file has deprecated options that aren't in the template anymore, confusing users, or if the template has options not in the real config file, so they'd have to add it to set it anyway.

ThreeDeeJay: Fair enough👌 
So this would be the minimal config to ensure HRTF, regardless of values set in the global config, right?
```ini
[general]
channels=stereo
stereo-mode=headphones
stereo-encoding=hrtf
```
tbf I had intentionally kept `hrtf` for backward compatibility in games that require older builds to work around some issues, like [Deus Ex with ALAudio](https://airtable.com/shrtTsUtiAJQ7wFGI/tblNOTdmp5nHXfFGU/viwjDPuKOjBnIQunv/recLK6rvfZs2a40IF) though I think [DSOAL](https://airtable.com/shrtTsUtiAJQ7wFGI/tblNOTdmp5nHXfFGU/viwjDPuKOjBnIQunv/recU6MI6CyiGQOQiW) works better with new builds anyway.

Also, I should point out that the main release package won't have hard-coded config for HRTF, only the template:
https://github.com/ThreeDeeJay/openal-soft/releases/download/latest/OpenALSoft.zip 
https://github.com/ThreeDeeJay/openal-soft/releases/download/latest/OpenALSoft+HRTF.zip as an separate file should make it obvious that the intent is to use HRTF, even if the device's set to surround (because I think the average person just uses the onboard sound and might not even realize if the system config is set to surround, so not forcing HRTF might actually cause missing audio channels on since only front left and front right would be heard on headphones)

> WASAPI shouldn't have an issue detecting when the device is set to headphones

wait, so is it possible to change the windows audio device type or do you mean via `stereo-mode`?
IME OpenAL Soft seems to always detect surround-capable devices as speakers (even when set to stereo), and only certain devices like the front panel, DACs (and ofc headphones with built-in DACs) as headphones.

> if it is headphones but is using virtual surround sound to simulate surround sound speakers, won't sound good 

Yeah, unfortunately that's one of the inconveniences we have to deal with. 
The only possible solutions I could think of are https://github.com/kcat/openal-soft/issues/586, https://github.com/kcat/openal-soft/issues/682 or somehow interfacing with a VSS like HeSuVi via EqualizerAPO, or replacing them altogether with a custom APO, though I guess that's a bit outside of the project's scope. 🤔

kcat: Sorry, I somehow missed replying.

> So this would be the minimal config to ensure HRTF, regardless of values set in the global config, right?

Yes.

> wait, so is it possible to change the windows audio device type or do you mean via `stereo-mode`?

I mean in Windows. It's been a while since I dealt with it, I don't use Windows often and I don't have access to a Windows system right now, but I remember there being a dropdown selection somewhere in the audio device configuration to say whether a stereo device is Speakers or Headphones, which correlated to the FormFactor property that OpenAL Soft's WASAPI backend reads to autodetect headphones.

If you're not seeing it, maybe it's only available for certain devices/drivers, or it was moved/removed in other versions of Windows, which would be unfortunate.

ThreeDeeJay: No worries. And I'm not aware of a setting to set device type on the Windows GUI side, usually only on the driver's control panel, but that doesn't seem to change how Windows itself treats the device. Like on my USB X-Fi, it just enables Headphone HRTF when CMSS-3D is enabled, as opposed to speaker HRTF otherwise IIRC, but Windows still treats it as a speaker.
image
AFAIK the headphones and stereo speaker configurations were merged into just Stereo after [Windows XP](https://developer.valvesoftware.com/wiki/Talk:Snd_surround_speakers). 
On a side note, have you considered letting some [games](https://i.imgur.com/cqGjf3Z.png)/[apps](https://imgur.com/lT2xGzh)' headphone setting override the system's when using DSOAL in case it can be changed?

Though I suspect the device type may be buried in a [registry device property key](https://github.com/AutomatedLab/DeviceManagement/blob/master/DeviceManagementLib/Device.cs#L113-L289) like `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\MMDevices\Audio\Render\DEVICE_GUID\Properties`, unless it's hard coded into the driver. Here's the registry values from my onboard Realtek front (Headphones) vs rear (Speakers) panel playback devices: [RealtekRegistryProperties.zip](https://github.com/user-attachments/files/16503195/RealtekRegistryProperties.zip)
image

My X-Fi only has a single headphone port, though I think it used to automatically switch to headphones when it was plugged in but I think that only work on an older Windows/driver. But I digress.

@mirh Are you familiar with the playback device type (headphones vs speaker) property on modern Windows?

kcat: > On a side note, have you considered letting some [games](https://i.imgur.com/cqGjf3Z.png)/[apps](https://imgur.com/lT2xGzh)' headphone setting override the system's when using DSOAL in case it can be changed?

Maybe. As far as I know, the app controls that with `IDirectSound::SetSpeakerConfig` (and `IDirectSound::GetSpeakerConfig`). They're no-ops since Windows Vista according to Wine, likely to prevent individual apps from changing the system's device configuration, which is what DSOAL does too. Theoretically I could maybe watch for it to be set to `DSSPEAKER_HEADPHONE` and enable HRTF for the device, and leave it at the default for everything else. I'd have a slight worry about it potentially confusing apps since the setting wouldn't persist between closing/reopening the device, though an app probably couldn't rely on it persisting anyway since other apps could change it too, so it may not be a big deal.

> @kcat By the way, is there anything else I should change on this PR? Besides resolving merge conflicts, of course 😅

No, I think it looks alright once the conflicts are fixed.

mirh: You can't really expect creative's drivers to be representative of how things should properly work 🙃
(and regardless, I remember even my Xonar has an "internal separation" between what it presents to the system and the one it knows is physically connected to the the card)
And we had already discussed speakers and headphones in https://github.com/kcat/dsoal/issues/63 (applications shouldn't interfere with system configuration, but of course there's always edge cases). 

Nevertheless the property is there underneath (even though I could swear there was also somewhere a dropdown menu that let you select it, as kcat said.. but I cannot find it now)

ThreeDeeJay: > No, I think it looks alright once the conflicts are fixed.

@kcat Done. And while we're at it, let me know if https://github.com/kcat/dsoal/pull/52 needs any more changes 👀 
I'm planning to add a separate DSOAL build with HRTF too, when I can fetch the config from OpenAL Soft upstream, but it should be good to go as it is now. 👌 

> I could swear there was also somewhere a dropdown menu that let you select it, as kcat said.. but I cannot find it now

@mirh Are you referring to this?

firefox_q2GAprN2UQ
Apparently that dialog only shows up for the front panel which is already detected as headphones, and it doesn't even respect the speaker mode

mirh: That, I still get it. 
And that includes the back panel too. 

But I just could remember this native windows control panel with the selection...

ThreeDeeJay: like this flyout dialog or like something deeper in the sound settings?
image

> That, I still get it.
And that includes the back panel too.

Do any of them change how Windows detects the device?
I wonder if some sound cards just switch like a driver mode to present/cloak itself as a "different" speakers or headphones device instead of telling Windows to change it on its end. 🤔

mirh:
> My only concern is the issues mentioned [here](https://www.pcgamingwiki.com/wiki/Glossary:Surround_sound#OpenAL) when not using the router:

I really don't know why Cammy8580 felt entitled to smartass their way like that.. Anyway, even though that catastrophism isn't remotely justified, this new behaviour doesn't sound legit either.
I mean, until the usual blablabla #757 I do suppose there are a few scenarios that are more likely to work this way.. But I don't know, still? Maybe what annoys me most is people reinventing the wheel after every half decade, and then people years later having to figure out why the old instructions were so oddly different (making this seems like hot voodoo).

> So does the router really provide any fixes/improvements compared to using the OpenAL Soft driver library directly?

Obviously not when there really isn't any routing to do whatsoever by now (in the sense that openal-soft is so much smoking everything else out of the water, that there's no point in changing/switching it except perhaps for testing).
Still if not any when you have to hold together both games that shipped their own router and those that didn't (nonchalantly, because at the end of the day there always was just ONE supposed method for things to work) I actually don't even see how this would be easier _overall_.

cammysorammy:
> I really don't know why Cammy8580 felt entitled to smartass their way like that.. Anyway, even though that catastrophism isn't remotely justified, this new behaviour doesn't sound legit either. I mean, until the usual blablabla #757 I do suppose there are a few scenarios that are more likely to work this way.. But I don't know, still? Maybe what annoys me most is people reinventing the wheel after every half decade, and then people years later having to figure out why the old instructions were so oddly different (making this seems like hot voodoo).

I wasn't trying to "smartass" my way into anything. I saw a complete mess on the PCGW page that included a lot of non-essential instructions as if they were essential to the use of OpenAL and decided to completely revise the section to the best of my ability for the sake of ease-of-use by people who needed to use the project.

I had included the mention of the router because Kane & Lynch 2 will _always_ crash if you attempt to place a renamed soft_oal.dll in the game's directory. I don't know why it does this, but it does. It's why I included that specific note because I assumed the behavior would have poured into other games that had done the same thing. It was to bypass the issue that I Drink Lava had proclaimed in his issue about needing to replace the file not only in the game directory, but also the DLL located in the System64 directory, which didn't even work for me when I had attempted it. It would've fixed not only his problem, but would've made OpenAL Soft global, and anyone who wanted to update it could've simply just transferred their soft_oal.dll into both system files whenever it updated and used their own alsoft.inis on a per-game basis if they needed to do things like fix games having inverted elevation.

Should I have made the instructions focused a bit more on a per-game basis? Probably, and I most likely will edit it now, but [you could've just went to my talk page](https://www.pcgamingwiki.com/wiki/User_talk:Cammy8580) and had let me know if you were so bothered, or even just edited it yourself and corrected me if you were _really_ that miffed instead of waiting for me to find your critique myself. PCGW is a community wiki and anyone could've edited any information under there that was inaccurate. I was under the impression that it was fine, and [3DJ had even had a conversation with me](https://discord.com/channels/165183350913499136/326048443040530432/1215487783401955378) about this in the PCGW discord back when I had first made the edit and assumed he must've simply forgot.

cammysorammy:
@mirh Yesterday I had worked with 3DJ on editing PCGW to accomodate specific information pertaining to the newer builds that were added in his merge with the latest compiled builds.

image

Hope this isn't "reinventing the wheel" again, but if there's any information you disagree with, you're more than welcome to edit what I have on here.

Issue #1010, opened 2024-06-27
Android library failing to build 


ThreeDeeJay:
Some extra context: https://github.com/kcat/openal-soft/issues/824
I put together a [GitHub actions workflow](https://github.com/ThreeDeeJay/openal-soft/blob/android/.github/workflows/ci.yml#L16-L17) to build the Android library (`libopenal.so`) and noticed that a commit is causing some compilation errors. After some trial and error, I managed to track down the culprit, from June 1st:
- https://github.com/kcat/openal-soft/commit/a0e238f54dbf57a2c88f2b45a0c49f6a13e0784c - Last commit to [successfully compile](https://github.com/ThreeDeeJay/openal-soft/actions/runs/9703301432/job/26781110464#step:2:7)
- https://github.com/kcat/openal-soft/commit/2719ce2c0aaf40bdf2db1c3e579b3b76e9cbcf14 - Commit where compilation started [failing](https://github.com/ThreeDeeJay/openal-soft/actions/runs/9703265481/job/26780989241#step:5:107)

ThreeDeeJay:
On a side note, I'm not sure if it's related, but the the one that did build crashes [sView](https://github.com/ThreeDeeJay/sview/releases/download/latest/sView.apk) with this error when enabling HRTF in the settings, unlike [Centurion96's build](https://github.com/kcat/openal-soft/issues/824#issuecomment-1932451795) which works, though I think it lacks embedded HRTF which might be a clue. 🤔 

Here's the log starting right before enabling HRTF:
```

D com.sview: PlayerBase::stop() from IPlayer
D AudioTrack: stop(186): called with 477126 frames delivered
D openal: [ALSOFT] (II) ALC_OUTPUT_MODE_SOFT = 6578
D openal: [ALSOFT] (II) ALC_HRTF_SOFT = 2
D openal: [ALSOFT] (II) Pre-reset: *Stereo, Float32, 48000hz, 960 / 2880 buffer
D com.sview: PlayerBase::PlayerBase()
D com.sview: TrackPlayerBase::TrackPlayerBase()
I libOpenSLES: Emulating old channel mask behavior (ignoring positional mask 0x3, using default mask 0x3 based on channel count of 2)
D openal: [ALSOFT] (II) Post-reset: Stereo, Int16, 48000hz, 960 / 2880 buffer
D openal: [ALSOFT] (II) Searching / for *.mhr
[...]

Issue #824, opened 2023-02-21
Compiling OpenAL `1.23.0` for Android.


Brahvim:
Hello,

I'd like to know how I can _(separately)_ compile OpenAL for Android with an OpenSL-ES or Oboe 'background'.

I'm sorry, but my knowledge of build systems is bleak.

All of my personal attempts ended with me generating a Visual Studio solution, and building that, which would give me DLLs and not Shared Objects.

I have also tried using the CMake commands mentioned in the [`XCompile-Android.txt`](https://github.com/kcat/openal-soft/blob/master/XCompile-Android.txt) file, but they throw errors about my Visual Studio installation not supporting Clang (even after installing the 'toolchain' components for Mobile Development in Visual Studio).

Could you please provide guidance on what I should be doing?

I thank you, in advance.

kcat:
You will need to install a compiler toolchain for building code for Android. The Android NDK provides compilers for compiling to Android from other systems. Unfortunately I don't know much about VS's Android support to know how to build a CMake project with it to generate Android binaries.

What exactly did you try to run, and what was the output? Can you paste it here?

Brahvim:
...I guess it's not a good idea to use Visual Studio for this right now.

I get the following output when using CMake (`cmake .. -DANDROID_STL=c++_shared -DCMAKE_TOOLCHAIN_FILE=C:\Projects\Android\AAA_LIBS\SDK\ndk\21.4.7075529\build\cmake\android.toolchain.cmake`):

```
-- Building for: Visual Studio 17 2022
-- ANDROID_PLATFORM not set. Defaulting to minimum supported version
16.
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - failed
-- Check for working C compiler: C:/Projects/Android/AAA_LIBS/SDK/ndk/21.4.7075529/toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe
-- Check for working C compiler: C:/Projects/Android/AAA_LIBS/SDK/ndk/21.4.7075529/toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe - broken
CMake Error at C:/Program Files/CMake/share/cmake-3.23/Modules/CMakeTestCCompiler.cmake:69 (message):
  The C compiler

    "C:/Projects/Android/AAA_LIBS/SDK/ndk/21.4.7075529/toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe"

  is not able to compile a simple test program.

[...]

PR #1025, opened 2024-08-03
Build for Android (32 and 64-bit) and upload as workflow artifacts

ThreeDeeJay:
Context: https://github.com/kcat/openal-soft/issues/1010 and https://github.com/kcat/openal-soft/issues/824
This builds [32-bit/armeabi-v7a](https://github.com/ThreeDeeJay/openal-soft/actions/runs/10224078457/artifacts/1771370955)  and [64-bit/arm64-v8a](https://github.com/ThreeDeeJay/openal-soft/actions/runs/10224078457/artifacts/1771371327) libopenal.so for Android that can replace older/generic OpenAL library from Android apps/games (if using root to access `/data/app/*hash*/*package*/lib/...`), or just include it in an android app/game rebuilt from source. Tested in [sView](https://youtu.be/ssmLgkbRMh0) (32-bit) and IdTech4A/Quake4 (64-bit).

Though apparently an update in android or something on my end broke apps' ability to load `/etc/openal/alsoftrc.config` (the only location that [worked for me to load external MHRs](https://github.com/kcat/openal-soft/issues/824#issuecomment-1937867413)) so I wonder if it would be possible to force HRTF on by default without relying on a config file since[ it doesn't even load the config or MHR from the supposed executable/working directory (`/system/bin/`](https://i.imgur.com/Mv0R1Qb.png) like it does on Windows. 🤔

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,7 @@ jobs:
         - {
             name: "Android_armeabi-v7a-Release",
             os: ubuntu-latest,
-            cmake_opts: "-DCMAKE_TOOLCHAIN_FILE=/usr/local/lib/android/sdk/ndk/26.3.11579264/build/cmake/android.toolchain.cmake \
+            cmake_opts: "-DCMAKE_TOOLCHAIN_FILE=$ANDROID_NDK_HOME/build/cmake/android.toolchain.cmake \
             -DALSOFT_EMBED_HRTF_DATA=TRUE \
             -DALSOFT_REQUIRE_OPENSL=ON",
             build_type: "Release"
@@ -119,7 +119,7 @@ jobs:
             os: ubuntu-latest,
             cmake_opts: "-DANDRIOD_ABI=arm64-v8a \
             -DANDROID_PLATFORM=25 \
-            -DCMAKE_TOOLCHAIN_FILE=/usr/local/lib/android/sdk/ndk/26.3.11579264/build/cmake/android.toolchain.cmake \
+            -DCMAKE_TOOLCHAIN_FILE=$ANDROID_NDK_HOME/build/cmake/android.toolchain.cmake \
             -DALSOFT_EMBED_HRTF_DATA=TRUE \
             -DALSOFT_REQUIRE_OPENSL=ON",
             build_type: "Release"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,14 +112,6 @@ jobs:
             cmake_opts: "-DCMAKE_TOOLCHAIN_FILE=/usr/local/lib/android/sdk/ndk/26.3.11579264/build/cmake/android.toolchain.cmake \
             -DALSOFT_EMBED_HRTF_DATA=TRUE \
             -DALSOFT_REQUIRE_OPENSL=ON",
-            deps_cmdline: "sudo apt update && sudo apt-get install -qq \
-              libpulse-dev \
-              portaudio19-dev \
-              libasound2-dev \
-              libjack-dev \
-              libpipewire-0.3-dev \
-              qtbase5-dev \
-              libdbus-1-dev",
             build_type: "Release"
           }
         - {
@@ -130,14 +122,6 @@ jobs:
             -DCMAKE_TOOLCHAIN_FILE=/usr/local/lib/android/sdk/ndk/26.3.11579264/build/cmake/android.toolchain.cmake \
             -DALSOFT_EMBED_HRTF_DATA=TRUE \
             -DALSOFT_REQUIRE_OPENSL=ON",
-            deps_cmdline: "sudo apt update && sudo apt-get install -qq \
-              libpulse-dev \
-              portaudio19-dev \
-              libasound2-dev \
-              libjack-dev \
-              libpipewire-0.3-dev \
-              qtbase5-dev \
-              libdbus-1-dev",
             build_type: "Release"
           }
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,40 @@ jobs:
               libdbus-1-dev",
             build_type: "Release"
           }
-
+        - {
+            name: "Android_armeabi-v7a-Release",
+            os: ubuntu-latest,
+            cmake_opts: "-DCMAKE_TOOLCHAIN_FILE=/usr/local/lib/android/sdk/ndk/26.3.11579264/build/cmake/android.toolchain.cmake \
+            -DALSOFT_EMBED_HRTF_DATA=TRUE \
+            -DALSOFT_REQUIRE_OPENSL=ON",
+            deps_cmdline: "sudo apt update && sudo apt-get install -qq \
+              libpulse-dev \
+              portaudio19-dev \
+              libasound2-dev \
+              libjack-dev \
+              libpipewire-0.3-dev \
+              qtbase5-dev \
+              libdbus-1-dev",
+            build_type: "Release"
+          }
+        - {
+            name: "Android_arm64-v8a-Release",
+            os: ubuntu-latest,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -106,7 +106,40 @@ jobs:
               libdbus-1-dev",
             build_type: "Release"
           }
-
+        - {
+            name: "Android_armeabi-v7a-Release",
+            os: ubuntu-latest,
+            cmake_opts: "-DCMAKE_TOOLCHAIN_FILE=/usr/local/lib/android/sdk/ndk/26.3.11579264/build/cmake/android.toolchain.cmake \
```
kcat, line 112: Out of curiosity, where does the SDK and version number come from? It doesn't look like anything's installing it, and that version looks like it's prone to change. And are any of the apt packages needed for the Android build?

kcat:
> Though apparently an update in android or something on my end broke apps' ability to load /etc/openal/alsoftrc.config (the only location that https://github.com/kcat/openal-soft/issues/824#issuecomment-1937867413) so I wonder if it would be possible to force HRTF on by default without relying on a config file since[ it doesn't even load the config or MHR from the supposed executable/working directory (/system/bin/](https://i.imgur.com/Mv0R1Qb.png) like it does on Windows. 🤔

HRTF being on by default would cause a problem for people using speakers, reducing the quality and increasing CPU load. If it can't detect the device as being headphones, then it needs either a config file or the app to enable it (or a modified build that uses HRTF by default).

.github/workflows/ci.yml
ThreeDeeJay: GitHub actions runners come with NDK pre-installed, but its directory variable wasn't defined so I had to hardcode the path after finding it by recursively listing all files in the drive. Kinda hacky, I know lol
But I'll see if I can find a way to set/get the proper variable for a cleaner/relative path. 👌 

As for the packages, I'm not sure if they're all needed. I just used the linux build settings as a template since that's technically what Android is based on and what the runner uses, but I'll try building without them, though I'm not sure it'd inadvertently break something so let me know if there's any packages that are probably needed for core functions.

kcat: I don't think any of the packages should be needed. I don't think they'd work anyway since they're for the system it's compiling on, which wouldn't work for the Android target (though it looks like it may be picking up the pipewire and dbus pkg-configs, which it shouldn't because Android doesn't have them as far as I know; luckily it doesn't cause a build failure since they're optional at runtime).

ThreeDeeJay: I just looked at the [log](https://github.com/user-attachments/files/16482376/log_2024_08_03_13_35_52.log) and both indeed fail to load/initialize so I'll go ahead and remove the whole `deps_cmdline` chunk and see how it goes. 🤞

ThreeDeeJay: btw just found the proper NDK path variable https://github.com/kcat/openal-soft/pull/1025/commits/6538bff88dcd7b0ecf6c47c8e2c46ce1d95d5721 👌

ThreeDeeJay:
Yeah, just like we were discussing on https://github.com/kcat/openal-soft/pull/1008 about Windows, it doesn't seem to be able to detect output device as headphones reliably (I know it's probably hard to tell whether the 3.5mm jack is being used for headphones or speakers, but in sView, I still need to force HRTF in the app settings even when using my bluetooth headhphones which are recognized as headphones by Android). 

So perhaps it might be handy to offer a separate HRTF build since that's probably the main reason why people would probably want to replace games/apps' OpenAL library. Or do you think it'd be possible to make OpenAL Soft load the config file from a [path accessible by the app, like its data folder/s](https://github.com/kcat/openal-soft/issues/824#issuecomment-1937867413)?
I just checked and it seems the stock file manager can read and write to <InternalStorage>/Android/data/packagename/ without needing root.

Also any idea if it would be possible to swap a compiled app's library on Android without having to root, like we do on Windows?
If I understand correctly the license requires being able to replace it, but most people don't bother rooting their phones so I wonder if there's an easier way 🤔

kcat:
> Yeah, just like we were discussing on #1008 about Windows, it doesn't seem to be able to detect output device as headphones reliably (I know it's probably hard to tell whether the 3.5mm jack is being used for headphones or speakers, but in sView, I still need to force HRTF in the app settings even when using my bluetooth headphones which are recognized as headphones by Android).

There's possibly a query to get the formfactor of the audio device that I don't know about, which could be added since the OpenSL and Oboe backends can't currently detect that. It might be an issue if the query can only be made via the Java API, instead of OpenSL, Oboe, or some other C or C++ API, since even though Java can be called from C/C++ with appropriate wrappers, it typically relies on some app handle given to the Java code (which at best would need a new extension function for the app to pass it in to OpenAL Soft, as it's not directly accessible).

> Or do you think it'd be possible to make OpenAL Soft load the config file from a [path accessible by the app, like its data folder/s](https://github.com/kcat/openal-soft/issues/824#issuecomment-1937867413)? I just checked and it seems the stock file manager can read and write to /Android/data/packagename/ without needing root.

Loading config files and data files from extra Android-specific paths is certainly an option, if there's some way to get that path (or at least the `packagename` to construct the path).

> Also any idea if it would be possible to swap a compiled app's library on Android without having to root, like we do on Windows? If I understand correctly the license requires being able to replace it, but most people don't bother rooting their phones so I wonder if there's an easier way 🤔

I don't think so. One of the complaints people have with phones is that they're more locked down than PCs, that they make it harder (though not impossible, and Android being less draconian) to access and replace files. At the very least, I think developer mode would need to be enabled to connect it to a PC as a storage device to access and transfer files, but it's not something I've messed around with very much.

ThreeDeeJay:
> There's possibly a query to get the formfactor of the audio device that I don't know about, which could be added since the OpenSL and Oboe backends can't currently detect that. It might be an issue if the query can only be made via the Java API, instead of OpenSL, Oboe, or some other C or C++ API, since even though Java can be called from C/C++ with appropriate wrappers, it typically relies on some app handle given to the Java code (which at best would need a new extension function for the app to pass it in to OpenAL Soft, as it's not directly accessible).

Perhaps something like [AudioDeviceInfo](https://developer.android.com/reference/android/media/AudioDeviceInfo)?
> Loading config files and data files from extra Android-specific paths is certainly an option, if there's some way to get that path (or at least the packagename to construct the path).

[getExternalFilesDirs](https://developer.android.com/reference/android/content/Context#getExternalFilesDirs(java.lang.String)) seems like what we need.

> I don't think so. One of the complaints people have with phones is that they're more locked down than PCs, that they make it harder (though not impossible, and Android being less draconian) to access and replace files. At the very least, I think developer mode would need to be enabled to connect it to a PC as a storage device to access and transfer files, but it's not something I've messed around with very much.

Ohh that's right. I've taken root for granted for so long that I keep completely forgetting (rooted) ADB is a thing.
[ADB Explorer](https://github.com/Alex4SSB/ADB-Explorer) is a fairly user-friendly way to modify system/app data files.
ADB_Explorer_EIK3ehfyeE
image
image
The only tricky part is finding the app folder hash after `/data/app/` which I only know how to get in [AppManager](https://github.com/MuntashirAkon/AppManager/releases) and of course we also need to know the package name.
Also there are apps that use a Wireless debugging loopback to run adb commands from the same device, though I'm not sure if they allow root or if there's a GUI/file manager that can do what ADB Explorer does. 🤔

kcat:
> Perhaps something like [AudioDeviceInfo](https://developer.android.com/reference/android/media/AudioDeviceInfo)?

Yeah, but I don't know how to get that info in C++. Doesn't even look like the AAudio API has anything for that (which Oboe sits on top of, alongside OpenSL).

> [getExternalFilesDirs](https://developer.android.com/reference/android/content/Context#getExternalFilesDirs(java.lang.String)) seems like what we need.

Possibly, but that's also through Java.

Issue #1015, opened 2024-07-08
[Question] What's the difference between regular releases an makemhr releases?


vanfanel:
Hi there,

Looking at the OpenAL-soft releases in https://github.com/kcat/openal-soft/releases, I see there are regular releases (latest seems to be version 1.23.1 from back in 2023) and mkhr releases.

What's the difference? Which should be used to build on GNU/Linux for performance and stability?

kcat:
The makemhr pre-releases are made when the `makemhr` utility has an update. Either the makemhr source itself was modified, or something it includes.

vanfanel:
I never thanked you guys for your answers on this, sorry!
Better late than never.... so many thanks! :)

Issue #999, opened 2024-05-27
[Question] 16 speakers in Windows using AmbiSonics (without HDMI and Receiver)


RennMekka:
Hello kCat, 

Are there any plans to use the full AmbiSonics 3rd order (16 speakers max) with OpenAl-Soft in a way that I can directly route 16 
 channels from the internal decoder to discrete outputs, let's say 16 ASIO channels on two sound cards?

I don't know how to explain it better, but I want to use up to 16 speakers with the outputs of two ASIO cards. (no Receiver, and also no HDMI)

The only way to move around the Windows 7.1 config (as I see it) would be to directly address the ASIO channels that are available in the system. 
So the decoder has to use an algorythm that produces true 16 channels from the internal B-stream, and map them to the channel-table. 

But I assume that OpenAL-Soft never support the tool to do the matrix calculations for 16 speakers and the ambdec config file itself.
The "highest" configuration at the moment uses 7.1.4 speakers. The downside of it is that is produces no sound from below. 

The idea was that I built my own matrix/rig of speakers (up to 16), but I don't know at the moment how to implement this with OpenAL-Soft.

kcat:
Hi.

The 16 channels for the 3rd order ambisonic mix that OpenAL Soft uses internally aren't 16 discrete outputs/speaker feeds. It needs to be decoded before being played back, and actually for a proper reproduction of a full 3rd order ambisonic mix, you need more than 16 speakers (you need more speakers than ambisonic channels to properly play it back) and those speakers need to be arranged in a suitable layout to sound good. The internal decoders OpenAL Soft has is for outputting to the supported channel configurations; mono, stereo, quad, 5.1, 6.1, and 7.1 (along with an experimental 7.1.4 output, and an uncommon 3D7.1 setup). It can also decode using HRTF (binaural/headphones), or convert to stereo-compatible UHJ. Alternatively, the ambisonic mix can be output directly without decoding or converting it, but that requires a suitable output API and an external decoder if it's to be listened to.

[...]

PR #1026, opened 2024-08-03
Upgrade makemhr workflow to build and upload utils

ThreeDeeJay:
- Includes makemhr, sofa-info, alrecord, altonegen, openal-info and allafplay Windows executables in [utils.zip](https://github.com/ThreeDeeJay/openal-soft/releases/download/utils/utils.zip)
- Uses my libmysofa fork that fixed the `Invalid format` error by removing a [filesize/density limit](https://github.com/hoene/libmysofa/issues/195#issuecomment-1370881871), so that makemhr is able to convert larger SOFA files (like [EAC](https://github.com/davircarvalho/Individualized_HRTF_Synthesis) HRTFs with 5° elevations/azimuths)
[v1.3.2](https://github.com/hoene/libmysofa/releases/tag/v1.3.2) (latest upstream)
image
[My fork](https://github.com/hoene/libmysofa/compare/main...ThreeDeeJay:libmysofa:main) (just a two-liner so let me know if you'd rather fork and remove limit yourself or if I should revert to the limited official branch)
image

I've had this branch for a while and was reminded of it [here](https://github.com/kcat/openal-soft/issues/1015#issuecomment-2238809390) and then I recently saw https://github.com/kcat/openal-soft/commit/3b10c6f9bc83e535f5de119f668ce6fd17e4e67d and wanted to try it out so I figured it'd be a good time to share this new workflow. Which by the way, is playback supposed to work or is it still a WIP? When I try the LAF files I posted [here](https://github.com/kcat/openal-soft/issues/999#issuecomment-2241620606), I just get a channel breakdown and it immediately exits the process without playing any audio 🤔 
image

### DIFF
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -30,7 +30,7 @@ jobs:
       run: |
         git clone https://github.com/ThreeDeeJay/libmysofa.git
         cd libmysofa
-        git checkout e72a8e1b06cab2c77b913e59f017667710c40ba9
+        git checkout 8642646ee6caca9085456bfa9d731200d68c25ca
 
     - name: Add MSBuild to PATH
       uses: microsoft/setup-msbuild@v1.1.3
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -1,10 +1,10 @@
-name: makemhr
+name: utils
 
 on:
   push:
     paths:
-      - 'utils/makemhr/**'
-      - '.github/workflows/makemhr.yml'
+      - 'utils/**'
+      - '.github/workflows/utils.yml'
 
   workflow_dispatch:
 
@@ -25,7 +25,7 @@ jobs:
       run: echo "CommitHash=$(git rev-parse --short=7 HEAD)" >> $env:GITHUB_ENV
 
     - name: Clone libmysofa
-      run: git clone --depth 1 --branch v1.3.1 https://github.com/hoene/libmysofa.git libmysofa
+      run: git clone --depth 1 https://github.com/ThreeDeeJay/libmysofa.git
 
     - name: Add MSBuild to PATH
       uses: microsoft/setup-msbuild@v1.1.3
@@ -52,25 +52,30 @@ jobs:
     - name: Collect artifacts
       run: |
         copy "build/Release/makemhr.exe" "Artifacts/makemhr.exe"
+        copy "build/Release/sofa-info.exe" "Artifacts/sofa-info.exe"
+        copy "build/Release/alrecord.exe" "Artifacts/alrecord.exe"
+        copy "build/Release/altonegen.exe" "Artifacts/altonegen.exe"
+        copy "build/Release/openal-info.exe" "Artifacts/openal-info.exe"
+        copy "build/Release/allafplay.exe" "Artifacts/allafplay.exe"
         copy "libmysofa/windows/third-party/zlib-1.2.11/bin/zlib.dll" "Artifacts/zlib.dll"
 
-    - name: Upload makemhr artifact
+    - name: Upload artifacts
       uses: actions/upload-artifact@v4
       with:
-        name: makemhr
+        name: utils
         path: "Artifacts/"
 
     - name: Compress artifacts
       uses: papeloto/action-zip@v1
       with:
         files: Artifacts/
-        dest: "Release/makemhr.zip"
+        dest: "Release/utils.zip"
 
     - name: GitHub pre-release
       uses: "marvinpinto/action-automatic-releases@latest"
       with:
         repo_token: "${{secrets.GITHUB_TOKEN}}"
-        automatic_release_tag: "makemhr"
+        automatic_release_tag: "utils"
         prerelease: true
-        title: "[${{env.CurrentDate}}] makemhr-${{env.CommitHash}}"
-        files: "Release/makemhr.zip"
+        title: "[${{env.CurrentDate}}] utils-${{env.CommitHash}}"
+        files: "Release/utils.zip"
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -27,6 +27,7 @@ jobs:
     - name: Clone libmysofa
       run: |
         git clone https://github.com/ThreeDeeJay/libmysofa.git
+        cd libmysofa
         git checkout e72a8e1b06cab2c77b913e59f017667710c40ba9
 
     - name: Add MSBuild to PATH
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -25,7 +25,9 @@ jobs:
       run: echo "CommitHash=$(git rev-parse --short=7 HEAD)" >> $env:GITHUB_ENV
 
     - name: Clone libmysofa
-      run: git clone --depth 1 --branch e72a8e1b06cab2c77b913e59f017667710c40ba9 https://github.com/ThreeDeeJay/libmysofa.git
+      run: |
+        git clone https://github.com/ThreeDeeJay/libmysofa.git
+        git checkout e72a8e1b06cab2c77b913e59f017667710c40ba9
 
     - name: Add MSBuild to PATH
       uses: microsoft/setup-msbuild@v1.1.3
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -25,7 +25,7 @@ jobs:
       run: echo "CommitHash=$(git rev-parse --short=7 HEAD)" >> $env:GITHUB_ENV
 
     - name: Clone libmysofa
-      run: git clone --depth 1 https://github.com/ThreeDeeJay/libmysofa.git
+      run: git clone --depth 1 --branch v1.3.2 https://github.com/hoene/libmysofa.git
 
     - name: Add MSBuild to PATH
       uses: microsoft/setup-msbuild@v1.1.3
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -25,7 +25,7 @@ jobs:
       run: echo "CommitHash=$(git rev-parse --short=7 HEAD)" >> $env:GITHUB_ENV
 
     - name: Clone libmysofa
-      run: git clone --depth 1 --branch v1.3.2 https://github.com/hoene/libmysofa.git
+      run: git clone --depth 1 --branch e72a8e1b06cab2c77b913e59f017667710c40ba9 https://github.com/ThreeDeeJay/libmysofa.git
 
     - name: Add MSBuild to PATH
       uses: microsoft/setup-msbuild@v1.1.3
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -10,6 +10,7 @@ on:
 
 env:
   BUILD_TYPE: Release
+  Branch: ${{github.ref_name}}
 
 jobs:
   Win64:
@@ -17,7 +18,7 @@ jobs:
 
     steps:
     - name: Clone repo and submodules
-      run: git clone https://github.com/${{github.repository}}.git .
+      run: git clone https://github.com/${{github.repository}}.git . --branch ${{env.Branch}}
 
     - name: Get current date, commit hash and count
       run: |
--- a/.github/workflows/utils.yml
+++ b/.github/workflows/utils.yml
@@ -24,7 +24,7 @@ jobs:
       run: |
         echo "CommitDate=$(git show -s --date=format:'%Y-%m-%d' --format=%cd)" >> $env:GITHUB_ENV
         echo "CommitHashShort=$(git rev-parse --short=7 HEAD)" >> $env:GITHUB_ENV
-        echo "CommitCount=$(git rev-list --count ${{env.Branch}})" >> $env:GITHUB_ENV
+        echo "CommitCount=$(git rev-list --count ${{env.Branch}} --)" >> $env:GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


kcat:
> * Includes makemhr, sofa-info, alrecord, altonegen, openal-info and allafplay Windows executables in [utils.zip](https://github.com/ThreeDeeJay/openal-soft/releases/download/utils/utils.zip)
> 
>     * Uses my libmysofa fork that fixed the `Invalid format` error by removing a [filesize/density limit](https://github.com/hoene/libmysofa/issues/195#issuecomment-1370881871), so that makemhr is able to convert larger SOFA files (like [EAC](https://github.com/davircarvalho/Individualized_HRTF_Synthesis) HRTFs with 5° elevations/azimuths)
>       [v1.3.2](https://github.com/hoene/libmysofa/releases/tag/v1.3.2) (latest upstream)
>       ...
>       [My fork](https://github.com/hoene/libmysofa/compare/main...ThreeDeeJay:libmysofa:main) (just a two-liner so let me know if you'd rather fork and remove limit yourself or if I should revert to the limited official branch)
>       ...

It's understandable to want to use a version that doesn't arbitrarily limit the supported size, that particular change seems questionable. It's removing checks for the sizes being negative, and with larger values, may need overflow checks in various other places. Improperly constructed SOFA files may be erroneously detected as valid, and cause misbehavior without overflow checks.

> Which by the way, is playback supposed to work or is it still a WIP? When I try the LAF files I posted [here](https://github.com/kcat/openal-soft/issues/999#issuecomment-2241620606), I just get a channel breakdown and it immediately exits the process without playing any audio 🤔 image

It should work, yes. Odd that it seems to fail reading the channel header data (it's all `0`s, when the `E` values should be `-0` or `-nan`), despite correctly reading the header markers and the quality, mode, and track count values, which in turn causes it to fail to read the sample rate and sample count, aborting playback. I don't see why, the file is opened in binary mode; if there was a problem reading binary data through `std::ifstream`, I'd have expected a similar problem with OpenAL Soft itself loading mhr files, which it obviously does without issue.

Unrelated, but it also seems like my trick of calling `std::terminate` in a `catch(...)` block doesn't seem to actually report anything on Windows? No dialog box about an abnormal termination, nor a message about why it terminated?

ThreeDeeJay:
> It's understandable to want to use a version that doesn't arbitrarily limit the supported size, but that particular change seems questionable. It's removing checks for the sizes being negative, and with larger values, may need overflow checks in various other places. Improperly constructed SOFA files may be erroneously detected as valid, and cause misbehavior without overflow checks.

Makes sense. I reverted it to the official repo though I updated it to v1.3.2 which has a slightly more generous limit.
So what if we just keep the negative size check, but expand the limit to around 4GB?
For reference a 1 degree azimuth/elevation EAC HRTF is >2GB so 96Khz might be double that, and I've seen other SOFAs that are GBs in size.
explorer_LceHBLufWm

> It should work, yes. Odd that it seems to fail reading the channel header data (it's all `0`s, when the `E` values should be `-0` or `-nan`), despite correctly reading the header markers and the quality, mode, and track count values, which in turn causes it to fail to read the sample rate and sample count, aborting playback. I don't see why, the file is opened in binary mode and those files play for me; if there was a problem reading binary/non-text data through `std::ifstream` on Windows, I'd have expected a similar problem with OpenAL Soft itself loading mhr files, which it obviously does without issue.

Looks like https://github.com/kcat/openal-soft/commit/7a7350af2b2cb9ad6103a43e7a932a4c45c31cb7 did the trick because this time I did get audio playback 👍 
image
Dolby's Universe Fury demo had never sounded this good to me with my HRTF, except, that there are really loud static noises at seemingly random times, which I just noticed also happen in [Cavern](https://github.com/VoidXH/Cavern/actions/runs/9965058277/artifacts/1708449527) when playing the [LAF](https://mega.nz/file/DexURLjY#weYtezaZ9CFiX0AVkBPQWrZ4sc0PHLr5oY2rhAmAch0), but when playing the [original file](https://web.archive.org/web/20230225220642/http://media.developer.dolby.com/Atmos/MP4/Universe_Fury2.mp4) directly, I instead get a suppressed pop/clip. So there might be a conversion bug or the Cavern decoder is yet to address some data getting corrupted 🤔 

> Unrelated, but it also seems like my trick of calling `std::terminate` in a `catch(...)` block doesn't seem to actually report anything on Windows? No dialog box about an abnormal termination, nor a message about why it terminated?

Nope, nada.

kcat:
> So what if we just keep the negative size check, but expand the limit to around 4GB? For reference a 1 degree azimuth/elevation EAC HRTF is >2GB so 96Khz might be double that, and I've seen other SOFAs that are GBs in size.

Looks like it may need an overflow check too. Something like
```c
if(elements <= 0 || size <= 0 || INT_MAX/size > elements)
    ... error ...
```

I don't know how this check in specific relates to the file or dataset size, but at least the result of `elements * size` can't go far above 2.1 billion, or else it'll overflow and become undefined behavior. Expanding that would need more work to avoid signed ints, where those kinds of changes tend to need farther reaching changes since those variables may interact with other variables that may not be happy when they're different types.

> Looks like [7a7350a](https://github.com/kcat/openal-soft/commit/7a7350af2b2cb9ad6103a43e7a932a4c45c31cb7) did the trick because this time I did get audio playback 👍

That's quite odd. It shouldn't have affected reading, basically just replacing terminating asserts with exceptions that skip just the current file.

> EDIT: Welp, apparently the static noise is a known issue with DD+ JOC [VoidXH/Cavern#17](https://github.com/VoidXH/Cavern/issues/17) So I wonder how [this person managed to get a clean sound](https://www.youtube.com/watch?v=N-sGfS7gT5E)
> 
> > This demo was converted from DD+ E-AC-3 JOC to LAF with Cavern and then rendered using OpenAL Soft's Default HRTF

I may not know what I'm talking about, or misunderstanding what I thought I read, but IIRC, support for pure DD+ JOC was a problem, while E-AC-3 JOC worked better.

> On the bright side, playing the [LAF](https://mega.nz/file/7fRnGCBI#eaJCMs8emSzxbrrMmR8HL_ZGKIIgsRUP6L1LFRlmlpc) converted from [ADM](https://discord.com/channels/547440061478862848/547441675115429889/1057700134659575910) ([invite](https://kutt.it/BinauralDiscord)) works beautifully

That's the kind of thing I expected to see from LAF files converted from Atmos. Rather than the other LAF file that had bursts of static, which seems to just be a 7.1.4 mix (that doesn't even mark the LFE channel properly), that one there uses a fixed channel bed (with a properly marked LFE channel) and several dozen dynamic objects. It does still seem to use left-handed coordinates though (+Z is front), which isn't correct according to the LAF documentation.

It does make me think the LAF player should try to be more efficient with source playback, stopping sources that are disabled in chunk headers, and starting them when enabled with more audio. Doing that would make it more efficient when multiple channels are disabled/silent most of the time. It would require the `AL_SOFT_source_start_delay` extension to make sure sources are restarted at the proper time synchronized with other sources (currently no extensions are required, outside of float32 and int32 for files that use float32 or int24 samples respectively), and if there's any resampling going on, synchronization will be extra tricky to account for sub-sample offsets.

ThreeDeeJay:
> Looks like it may need an overflow check too.

Done https://github.com/ThreeDeeJay/openal-soft/commit/f3d74878d2ac3e4e45b853569bec406dcb527bb3, though the [=<5° EAC SOFA](https://www.dropbox.com/s/ow1f3t46tw4tuvr/makemhr-f176484%2BEAC.SOFA-05ac8cc_InvalidFormat.7z?dl=1) still [fails to convert with the limit](https://github.com/hoene/libmysofa/issues/195#issuecomment-1382665204).
Also, the CI now clones a specific commit, for stability reasons, and I added commit date, count and hash to the workflow artifact to avoid duplicate confusion when debugging


> I may not know what I'm talking about, or misunderstanding what I thought I read, but IIRC, support for pure DD+ JOC was a problem, while E-AC-3 JOC worked better. Or maybe that was TrueHD with Atmos that didn't work, while DD+/E-AC-3 with Atmos did.

Yeah, what I do know is that [TrueHD can't be implemented at the moment](https://github.com/VoidXH/Cavern/issues/166#issuecomment-2053605565). 
So I've been looking into the process to at least [mix to 9.1.6](https://drive.google.com/drive/folders/175XNDo8C0Vth0tJFS2K5jErmKpJospOu?usp=sharing) so maybe I can convert it to ADM then into LAF, since source ADM tracks are extremely scarce. And apparently coarse JOC is what causes decoding artifacts:

> There are two types of JOC coding, coarse and sparse. Coarse, which is widely used by movies, works without issues, sparse (found in some versions of amaze) doesn't, as the documentation is incorrect about it
https://github.com/VoidXH/Cavern/issues/17#issuecomment-1268149518

kcat:
Wait, sorry. That check should be

```c
if(elements <= 0 || size <= 0 || elements > INT_MAX/size)
```

ThreeDeeJay:
Thanks, that did the trick! 👌 
By the way why does it sometimes use less than the total IRs? 🤔 
image

kcat:
> By the way why does it sometimes use less than the total IRs? 🤔

It ignores IRs that don't align with the detected uniform layout. The IR layout for the mhr file needs to have fixed stepping along the elevation (starting at +90 degrees and stepping down to -90 degrees), and fixed stepping along the azimuth for each valid elevation (starting at 0 degrees, directly in front, and going around a full 360 degrees).

ThreeDeeJay:
Nice 👍 
By the way, you might wanna delete https://github.com/kcat/openal-soft/releases/tag/makemhr since that release is now obsolete.

kcat:
> > I may not know what I'm talking about, or misunderstanding what I thought I read, but IIRC, support for pure DD+ JOC was a problem, while E-AC-3 JOC worked better. Or maybe that was TrueHD with Atmos that didn't work, while DD+/E-AC-3 with Atmos did.
> 
> Yeah, what I do know is that [TrueHD can't be implemented at the moment](https://github.com/VoidXH/Cavern/issues/166#issuecomment-2053605565). So I've been looking into the process to at least [mix to 9.1.6](https://drive.google.com/drive/folders/175XNDo8C0Vth0tJFS2K5jErmKpJospOu?usp=sharing) so maybe I can convert it to ADM then into LAF, since source ADM tracks are extremely scarce. And apparently coarse JOC is what causes decoding artifacts:
> 
> > There are two types of JOC coding, coarse and sparse. Coarse, which is widely used by movies, works without issues, sparse (found in some versions of amaze) doesn't, as the documentation is incorrect about it
> > [VoidXH/Cavern#17 (comment)](https://github.com/VoidXH/Cavern/issues/17#issuecomment-1268149518)

I'm curious if ffmpeg's code could help at all. It seems to handle decoding TrueHD tracks just fine, "truehd (Dolby TrueHD + Dolby Atmos)" even. Interestingly, it reports and provides such tracks as 7.1, which seem to have everything in the mix. I can't do a deeper look to see if it's 7.1(.4) premixed or something that it's giving with a simple downmix, or if it has dynamic objects being moved mixed in real-time by its decoder.

ThreeDeeJay:
> I'm curious if ffmpeg's code could help at all. It seems to handle decoding TrueHD tracks just fine, "truehd (Dolby TrueHD + Dolby Atmos)" even. "eac3 (Dolby Digital Plus + Dolby Atmos)" also plays fine, without any glitching. Interestingly, it reports and provides such tracks as 7.1, which seem to have everything in the mix. I can't do a deeper look to see if it's 7.1(.4) premixed or something that it's giving with a simple downmix, or if it has dynamic objects being moved and mixed in real-time by its decoder.

Last I checked, ffmpeg could just mux/identify the TrueHD Atmos track and decode the 7.1 "core" surround, but not heights or objects (which I think are encoded into the 7.1 track with metadata), though some figured out how to render up to 9.1.6 using a [script](<https://www.avsforum.com/threads/decoding-truehd-atmos-to-16-channel-wav.3265364/>) or [Music Media Helper](<https://gist.github.com/Plazik/e950267cc580bcde371ea7879aaa3d5d?permalink_comment_id=4822808#gistcomment-4822808>), both of which rely on Dolby Reference Player/Encoder


According to a user in our server, DDPA was implementable in FOSS since there is an open spec (ETSI EAC object carriage). Decoding THDA is not implementable in FOSS since there is no open spec, unless you get the proprietary software DRP to do the job, which is what the script does.
https://cavern.sbence.hu/cavern/doc.php?p=Cavernize#cd click on "Codec disclaimer" 

> Also interestingly, this LAF doesn't seem too well optimized. Each chunk always has all tracks enabled, when I'm sure many of them are silent for long stretches of time. I also wonder if there's room to optimize track utilization (if two or more tracks don't play sound at the same time, they can be on the same track; you'd only need as many tracks as are played simultaneously). Those would help cut down on the LAF file size pretty well, I think.

I forwarded it to the dev since tbh I'm not really familiar with the Cavern codebase (I'm just the CI guy there too lol), so feel free to join the [conversation here](https://github.com/VoidXH/Cavern/issues/186), or via a new issue/PR 👀👌

VoidXH:
> Also interestingly, this LAF doesn't seem too well optimized. Each chunk always has all tracks enabled, when I'm sure many of them are silent for long stretches of time. I also wonder if there's room to optimize track utilization (if two or more tracks don't play sound at the same time, they can be on the same track; you'd only need as many tracks as are played simultaneously). Those would help cut down on the LAF file size pretty well, I think.

When a LAF is exported by converting an E-AC-3 file, even if noise is present, the track will be active. One track is one object from the source. You could optimize it by clustering or gating the PCM signals, but that wouldn't be lossless.

PR #1029, opened 2024-08-08
Remove "v" from commit hash in release title

ThreeDeeJay:
Fixes a minor typo from https://github.com/kcat/openal-soft/pull/1008
image


ThreeDeeJay:
I could also include the branch to be consistent with the version printed to the log, which might be useful to identify forks/different branches, but I'm not sure if it'd be more redundant than useful.
image

Issue #1030, opened 2024-08-13
New stable release?


dustContributor:
Hi! I've been wondering, is it time for a new stable release? It's been a year and half since the previous one and there have been a looot of changes since then (~800 commits according to github,  "Showing with 35,430 additions and 27,137 deletions"). In particular I'm very interested in the debug extension that brings ARB_debug_output-like functionality to OpenAL.

In particular I'm asking because LWJGL doesn't picks up non-stable releases for their bindings so I'm "stuck" with the version from april 2023.

dustContributor:
Sorry I meant april 2023.

kcat:
Yes, I've been meaning to start getting a release together. But feature creep keeps creeping in, like the recent work with supporting ASIO drivers. I should probably put a hold on that to do afterward, though.

dustContributor:
Sweet, I hope you can make it.

kcat:
OpenAL Soft 1.24.0 has been released now., available on [the release page](https://github.com/kcat/openal-soft/releases/tag/1.24.0).

PR #1033, opened 2024-08-18
Enable Jack on Windows

ThreeDeeJay:
Following [this suggestion to try Jack on Windows](https://github.com/kcat/openal-soft/issues/999#issuecomment-2294123536), I decided to give it a shot, but found out that just setting `drivers=jack` wasn't enough: the backend was not enabled for Windows builds. So I managed to install the dependencies and build it here.
https://github.com/ThreeDeeJay/openal-soft/actions/runs/10440094213

However, I was still not able to get it working. I'm not sure if it's an issue with OpenAL Soft, Jack, PortAudio, ASIO, or something else on my end, but it can't even establish an active connection/stream via the Jack control panel (which btw is unnervingly buggy for an app that deals with ASIO 👀💦 )
image
Changing buffer or sample rate doesn't help, though fwiw the [dummy backend in Jack seems to play nice with ASIO4ALL](https://github.com/user-attachments/assets/728bdac7-01b6-47b5-8331-3e659048cae9) (though 32000hz is the highest sample rate without ["XRuns"](https://i.imgur.com/xG9rhrC.png) (underruns?) warnings)

Anyhow, when attempting to use Jack in OpenAL Soft, the apps fail to open the device no matter what I do:
ALCapsViewer32_CwwUNS7JE7

Test build with config and logs: [OpenALSoftJackTest.zip](https://github.com/user-attachments/files/16648235/OpenALSoftJackTest.zip)

Something to note: for some reason, the [Jack installers](https://github.com/jackaudio/jack2-releases/releases) copy DLLs to `C:\Windows` instead of `C:\Windows\System32`/`C:\Windows\SysWOW64` so I had to copy them manually to the executable folder. Also, OpenAL Soft Win64 looks for `libjack.dll` even though the file added by Jack is `libjack64.dll`.

I know this sounds more like an issue but even though I'm not really fond of the friction it takes to set this up, I figured I'd still share the CI in case someone can figure out how to get this working and maybe then this could be [eventually](https://github.com/kcat/openal-soft/issues/1030#issuecomment-2287727324) merged so people don't have to go out of their way to compile with Jack.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,12 @@ jobs:
               -DALSOFT_BUILD_ROUTER=ON \
               -DALSOFT_REQUIRE_WINMM=ON \
               -DALSOFT_REQUIRE_DSOUND=ON \
-              -DALSOFT_REQUIRE_WASAPI=ON",
-            build_type: "Release"
+              -DALSOFT_REQUIRE_WASAPI=ON \
+              -DALSOFT_REQUIRE_JACK=ON \
+              -DJACK_INCLUDE_DIR=\"C:\\Program Files (x86)\\JACK2\\include\" \
+              -DJACK_LIBRARY=\"C:\\Program Files\\JACK2\\lib32\\libjack.lib\"",
+            build_type: "Release",
+            architecture: "Win32"
           }
         - {
             name: "Win32-Debug",
@@ -29,8 +33,12 @@ jobs:
               -DALSOFT_BUILD_ROUTER=ON \
               -DALSOFT_REQUIRE_WINMM=ON \
               -DALSOFT_REQUIRE_DSOUND=ON \
-              -DALSOFT_REQUIRE_WASAPI=ON",
-            build_type: "Debug"
+              -DALSOFT_REQUIRE_WASAPI=ON \
+              -DALSOFT_REQUIRE_JACK=ON \
+              -DJACK_INCLUDE_DIR=\"C:\\Program Files (x86)\\JACK2\\include\" \
+              -DJACK_LIBRARY=\"C:\\Program Files\\JACK2\\lib32\\libjack.lib\"",
+            build_type: "Debug",
+            architecture: "Win32"
           }
         - {
             name: "Win64-Release",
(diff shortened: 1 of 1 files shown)


kcat:
Hmm, JACK seems to work here on Wine. There was a bit of misbehavior when I tried changing the Interface to certain devices, but I think that was more a quirk of Wine/winepulse.drv and PortAudio's requirements for full duplex mode. Using WineASIO works fine, and I could also go to the Advanced tab to set `Audio` to `Playback only`  from `Duplex` to make those other devices work. This is a pretty hefty audio stack (Windows JACK -> PortAudio -> WineASIO -> Linux JACK -> PipeWire), but I'm not using anything extreme (48khz with Frames/Period set to 512). and don't get any xruns (underruns or overruns).

And for OpenAL Soft, as long as the server was successfully running after pressing Start, it works fine with JACK for me. Fixing the backend to use `libjack64.dll` on 64-bit Windows was all I needed to do there (I didn't need to move the libjack DLLs from `C:\Windows`; `LoadLibrary` seems to pick them up fine from there).

I do notice something in your JACK Messages log that's different from me, though. Where it mentions `jack_client_close`, I see:
```
jack_client_close called with a NULL client
JackWinMMEDriver::Open - while creating input port: The driver was not enabled.
JackWinMMEDriver::Open - while creating input port: The driver was not enabled.
21:19:45.427 JACK connection change.
``` 
but you have a different message with `JackWinMMEDriver::Open`. Have you tweaked any JACK settings besides what you showed? Does it work if you turn up the `Frames/Period` setting?

ThreeDeeJay:
IIRC I had only changed interface, sample rate and periods (which I also tried maxing out with no luck), but after resetting everything in QJack Ctl to defaults, just setting Audio: `Playback only` seemed to allow starting some interfaces like FlexASIO (set to WASAPI), but even then, it completely ignores it along with sample rate and periods 
image
It also ignores the [*separate*  output device selection](https://i.imgur.com/QNnQwla.png).

Setting Server prefix to `jackd` (without `-S -X winmme`) stops it from loading ASIO4ALL when not selected, but when it *is* selected, [it still fails to start, though at least it respects the sample rate and periods](https://i.imgur.com/9zTIub2.png).
That setting (along with `Playback only`) also allows interfaces like MME, DirectSound and WASAPI to start and some work with OpenAL Soft, but it's still pretty buggy with errors, hangs and crashes often. 😔 

I expected this setup to be tedious, but I thought Jack wasn't supposed to be *this* clunky on Windows, unless again, I'm just missing something. We'd probably need a guide to set this up right, or just keep Jack away from windows if the ASIO/WASAPI exclusive backends are gonna be a thing eventually 👀👌

xximwon (changes requested):
H

PR #1038, opened 2024-08-27
Fix for Steam Deck crashing on loading the OpenALSoft library.

threax:
This PR fixes an issue where the library will not load on the Steam Deck. When I try to run my game with a newer version of OpenALSoft from master I get the error "0x not found in file . (0x8007013D)". The fix described below will fix this issue and allow it to run. Everything is fine on Windows.

This PR is simply a revert of the commit 33cd77b81b1dce9a5b55ec7e215315c500f9d0bf (* Use the C++ standard's regular modified Bessel function). However, if you decided to take this code you might want to review it, since I have no idea what it should do and I had to fix a couple of merge conflicts.

I'm not sure if this is something that should be fixed at this level, since you should be able to use the standard c++ functions, but maybe there is an issue with the implementation of them on Proton or something like that. Not an issue for this project to fix, but if someone else is like me and wants the latest version reverting that commit will make it work.

In my case I wanted to make use of the events to reload the audio device when it changes. To get the full api for that I have to go past the version that has this breaking change on the Deck.

### DIFF
--- a/common/polyphase_resampler.cpp
+++ b/common/polyphase_resampler.cpp
@@ -71,6 +71,33 @@ double Sinc(const double x)
     return std::sin(al::numbers::pi*x) / (al::numbers::pi*x);
 }
 
+/* The zero-order modified Bessel function of the first kind, used for the
+ * Kaiser window.
+ *
+ *   I_0(x) = sum_{k=0}^inf (1 / k!)^2 (x / 2)^(2 k)
+ *          = sum_{k=0}^inf ((x / 2)^k / k!)^2
+ */
+constexpr double BesselI_0(const double x)
+{
+    // Start at k=1 since k=0 is trivial.
+    const double x2{x/2.0};
+    double term{1.0};
+    double sum{1.0};
+    int k{1};
+
+    // Let the integration converge until the term of the sum is no longer
+    // significant.
+    double last_sum{};
+    do {
+        const double y{x2 / k};
+        ++k;
+        last_sum = sum;
+        term *= y * y;
+        sum += term;
+    } while(sum != last_sum);
+    return sum;
+}
+
 /* Calculate a Kaiser window from the given beta value and a normalized k
  * [-1, 1].
  *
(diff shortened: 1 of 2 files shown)


kcat:
Commit 326133368c3d7149d0ab4a98fc96f654553094a7 now avoids calling the standard `std::cyl_bessel_i` function. Can you test to make sure it fixes the issue? In either case, it may be worth reporting the issue to the Steam Deck people, if it's reporting the function to exist but then fail to load. And that error message seems broken, like it's missing some parts to it.

threax:
Hello. I tried your fix and it works as well.

When trying the broken version with proton logging the error it prints out is `err:module:loader_init "OpenAL32.dll" failed to initialize, aborting`. This made me think that its really an issue loading the library, not using it.

I investigated this a bit more and created [a repo](https://github.com/threax/OpenAlTest) with what I found. The short version is the use of `std::cyl_bessel_i` at all in a program will make it fail on the Steam Deck, since Wine apparently does not implement this function. I was able to pull this error from a proton log.

```
wine: Call from XXXXX to unimplemented function msvcp140_2.dll.__std_smf_cyl+bessel_i, aborting.
```

This came from running the StdCylBesselTest program in this repo. That one doesn't even link to OpenALSoft and has the same issue. I even found [another project](https://www.reddit.com/r/GraphicsProgramming/comments/1d9ewi0/how_i_learned_vulkan_and_wrote_a_small_game/) that is having this issue.

Given all of that information I think it makes sense to keep your fix and just use the functions you wrote. Maybe this can be revisited in the future when Wine adds support for this function.

threax:
I went ahead and added a post about this to the [WineHQ forums](https://forum.winehq.org/viewtopic.php?t=39367). Maybe someone over there has some more info. I just created my account, so a moderator has to approve the post if you can't see it right away.

ThreeDeeJay:
Neat, https://github.com/kcat/openal-soft/commit/326133368c3d7149d0ab4a98fc96f654553094a7 also fixed DSOAL crashing in [Winlator](https://github.com/brunodev85/winlator) for EAX/3D audio for DirectSound3D games on Android 👀👌

mirh:
They completely dropped the ball in [wine 6.22](https://bugs.winehq.org/show_bug.cgi?id=40898#c8)

Speaking instead of this issue, it seems stupid to workaround a [wine bug](https://bugs.winehq.org/show_bug.cgi?id=56999) (which I'm doubtful could even be called as such, considering that windows ***does NOT*** ship these libraries so it's a bit rich to pretend they should be reimplemented)

PR #1047, opened 2024-09-29
Add options to customize target names

michaelheilmann:


kcat:
Commit 76e4ebb81115257b806e89e53c3a39a7821f972d changes some of the internal target names to use the `alsoft.` prefix to avoid clashes with external projects. The remaining target names are either the name of the "public" binary produced or the expected external target name.

michaelheilmann:
Thank you, that seems to work as well. I will close the pull request.

PR #1056, opened 2024-11-04
Adding channel detection to CoreAudio backend

colincornaby:
Hi! I work on the a game engine that makes use of OpenAL Soft. We noticed, on macOS, that OpenAL was not properly providing surround sound to some hardware devices like the Apple Studio Display. The Studio Display is a 5.1.2 device, but OpenAL was treating it as a stereo device.

After some research - I found the CoreAudio backend for OpenAL Soft is not properly querying the channel count of devices. If the device type was overridden in the OpenAL config - surround output would work. But OpenAL would not auto detect surround devices on macOS.

This change implements a correct query of the device channel count by querying the preferred channel setup of the underlying device/audio unit. I took some notes from how PulseAudio handles this - and I also populate FmtChans by comparing against a known set of channel configurations. This change was implemented in two places. The function that gets the channel count is routed to the correct CoreAudio API. And a new chunk of code to populate FmtChans was added.

kAudioDevicePropertyPreferredChannelLayout also gives a channel layout for the audio buffers. I've opted to continue using the WaveFormatExtensible - but kAudioDevicePropertyPreferredChannelLayout can be used for manually populating the channel layout. One issue with using AudioChannelLayout to manually layout the OpenAL output channels is that it might be difficult to map from a manually specified OpenAL channel configuration back to CoreAudio layouts. It could be a possible improvement in the future.

This is my first PR to OpenAL - and my CoreAudio is a bit rusty. I'd appreciate any feedback on this PR! Thanks.

### DIFF
--- a/alc/backends/coreaudio.cpp
+++ b/alc/backends/coreaudio.cpp
@@ -56,29 +56,30 @@ namespace {
 constexpr auto OutputElement = 0;
 constexpr auto InputElement = 1;
 
-std::vector<AudioChannelLabel> MonoChanMap { kAudioChannelLabel_Mono };
-std::vector<AudioChannelLabel>  StereoChanMap { kAudioChannelLabel_Left, kAudioChannelLabel_Right};
-std::vector<AudioChannelLabel>  QuadChanMap {
+// These following arrays should always be defined in ascending AudioChannelLabel value order
+static constexpr std::array<AudioChannelLabel, 1> MonoChanMap { kAudioChannelLabel_Mono };
+static constexpr std::array<AudioChannelLabel, 2> StereoChanMap { kAudioChannelLabel_Left, kAudioChannelLabel_Right};
+static constexpr std::array<AudioChannelLabel, 4> QuadChanMap {
         kAudioChannelLabel_Left, kAudioChannelLabel_Right,
         kAudioChannelLabel_LeftSurround, kAudioChannelLabel_RightSurround
 };
-std::vector<AudioChannelLabel>  X51ChanMap {
+static constexpr std::array<AudioChannelLabel, 6> X51ChanMap {
         kAudioChannelLabel_Left, kAudioChannelLabel_Right,
         kAudioChannelLabel_Center, kAudioChannelLabel_LFEScreen,
         kAudioChannelLabel_LeftSurround, kAudioChannelLabel_RightSurround
 };
-std::vector<AudioChannelLabel>  X51RearChanMap {
+static constexpr std::array<AudioChannelLabel, 6> X51RearChanMap {
         kAudioChannelLabel_Left, kAudioChannelLabel_Right,
         kAudioChannelLabel_Center, kAudioChannelLabel_LFEScreen,
         kAudioChannelLabel_RearSurroundRight, kAudioChannelLabel_RearSurroundLeft
 };
-std::vector<AudioChannelLabel>  X61ChanMap {
+static constexpr std::array<AudioChannelLabel, 7> X61ChanMap {
         kAudioChannelLabel_Left, kAudioChannelLabel_Right,
         kAudioChannelLabel_Center, kAudioChannelLabel_LFEScreen,
         kAudioChannelLabel_CenterSurround,
         kAudioChannelLabel_RearSurroundRight, kAudioChannelLabel_RearSurroundLeft
 };
-std::vector<AudioChannelLabel>  X71ChanMap {
+static constexpr std::array<AudioChannelLabel, 8> X71ChanMap {
         kAudioChannelLabel_Left, kAudioChannelLabel_Right,
         kAudioChannelLabel_Center, kAudioChannelLabel_LFEScreen,
         kAudioChannelLabel_LeftSurround, kAudioChannelLabel_RightSurround,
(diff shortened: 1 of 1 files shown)


alc/backends/coreaudio.cpp
```diff
@@ -56,6 +56,35 @@ namespace {
...
+        kAudioChannelLabel_CenterSurround,
+        kAudioChannelLabel_RearSurroundRight, kAudioChannelLabel_RearSurroundLeft
+};
+std::vector<AudioChannelLabel>  X71ChanMap {
+        kAudioChannelLabel_Left, kAudioChannelLabel_Right,
+        kAudioChannelLabel_Center, kAudioChannelLabel_LFEScreen,
+        kAudioChannelLabel_LeftSurround, kAudioChannelLabel_RightSurround,
+        kAudioChannelLabel_LeftCenter, kAudioChannelLabel_RightCenter
+};
+
```
kcat, line 87: These should be `constexpr std::array`.

alc/backends/coreaudio.cpp
```diff
@@ -505,6 +530,22 @@ bool CoreAudioPlayback::reset()
         mDevice->Frequency = static_cast<uint>(streamFormat.mSampleRate);
     }
 
+    struct ChannelMap {
+        DevFmtChannels fmt;
+        std::vector<AudioChannelLabel> map;
+        bool is_51rear;
+    };
+    
+    static std::array<ChannelMap,7> chanmaps{{
```
kcat, line 539: `const`.

alc/backends/coreaudio.cpp
```diff
@@ -505,6 +530,22 @@ bool CoreAudioPlayback::reset()
         mDevice->Frequency = static_cast<uint>(streamFormat.mSampleRate);
     }
 
+    struct ChannelMap {
+        DevFmtChannels fmt;
+        std::vector<AudioChannelLabel> map;
```
kcat, line 535: `al::span`.

alc/backends/coreaudio.cpp
```diff
@@ -552,6 +593,44 @@ bool CoreAudioPlayback::reset()
...
+    {
+        UInt32 propSize{};
+        Boolean writable;
+        err = AudioUnitGetPropertyInfo(mAudioUnit,
+                                       kAudioUnitProperty_AudioChannelLayout,
+                                       kAudioUnitScope_Output,
+                                       OutputElement,
+                                       &propSize,
+                                       &writable);
+        AudioChannelLayout *layout = new AudioChannelLayout[propSize]();
```
kcat, line 607: This should also use `std::make_unique` or `std::vector` to avoid leaking the memory, and needs to use a raw bytes to reinterpret as the layout struct, like how `GetDeviceChannelCount` does it.

kcat (changes requested):

alc/backends/coreaudio.cpp
```diff
@@ -552,6 +593,44 @@ bool CoreAudioPlayback::reset()
...
+            for(UInt32 i=0; i<layout->mNumberChannelDescriptions; i++) {
+                tagsArray[i] = layout->mChannelDescriptions[i].mChannelLabel;
+            }
+            
+            if(err == noErr)
+            {
+                auto chaniter = std::find_if(chanmaps.cbegin(), chanmaps.cend(),
+                                             [&tagsArray](const ChannelMap &chanmap) -> bool
+                                             {
+                    return std::includes(tagsArray.begin(), tagsArray.end(), chanmap.map.begin(), chanmap.map.end());
```
kcat, line 624: I don't think this works as intended. According to [cppreference](https://en.cppreference.com/w/cpp/algorithm/includes), the arrays need to be sorted:

> Returns `true` if the sorted range [`first2`, `last2`) is a [subsequence](https://en.wikipedia.org/wiki/subsequence) of the sorted range [`first1`, `last1`) (a subsequence need not be contiguous).
> 1) If [`first1`, `last1`) or [`first2`, `last2`) is not [sorted](https://en.cppreference.com/w/cpp/algorithm#Requirements) with respect to operator<(until C++20) [std::less](http://en.cppreference.com/w/cpp/utility/functional/less){}(since C++20), the behavior is undefined.

But I don't think the given enum values are guaranteed to be ordered.

colincornaby: I've solved this by sorting the channel map that comes back from CoreAudio. (I don't believe Apple defines any particular order even though they seem to come back in ascending order.)

I also left a comment next to the constant values that any new values should be defined in ascending order. Probably a good tradeoff to prevent needing to sort those at runtime as well.

kcat:
Seems this is failing on iOS: https://github.com/kcat/openal-soft/actions/runs/11678208331/job/32535511458?pr=1056#step:6:397
```
error: use of undeclared identifier 'kAudioDevicePropertyPreferredChannelLayout'; did you mean 'kAudioFilePropertyChannelLayout'?
            err = AudioUnitGetProperty(mAudioUnit, kAudioDevicePropertyPreferredChannelLayout, kAudioUnitScope_Output, OutputElement, layout.get(), &propSize);
                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                   kAudioFilePropertyChannelLayout
```

The code block that tries to set the `DevChans` field might need to be conditioned on the `#if CAN_ENUMERATE` macro, unless the suggested change or something else can make it work.

kcat (approved):

PR #1084, opened 2024-12-15
Allow WASAPI exclusive mode to work in more devices

LAGonauta:
In all my devices exclusive mode wasn't working, and that is because `GetCurrentPadding` was always returning zero. This PR fixes in all my devices in my case, and there are no skipping or glitches.

`sample-type=int16` is still required for my devices, but I am going to resolve it later.

### DIFF
--- a/alc/backends/wasapi.cpp
+++ b/alc/backends/wasapi.cpp
@@ -1232,28 +1232,49 @@ FORCE_ALIGN int WasapiPlayback::mixerProc()
     auto avhandle = AvrtHandlePtr{AvSetMmThreadCharacteristicsW(taskname, &sAvIndex)};
 #endif
 
+    const auto sharemode =
+        GetConfigValueBool(mDevice->mDeviceName, "wasapi", "exclusive-mode", false)
+        ? AUDCLNT_SHAREMODE_EXCLUSIVE : AUDCLNT_SHAREMODE_SHARED;
+
     mBufferFilled = 0;
     while(!mKillNow.load(std::memory_order_relaxed))
     {
+        DWORD res{ WaitForSingleObjectEx(mNotifyEvent, 2000, FALSE) };
+        if (res != WAIT_OBJECT_0)
+            ERR("WaitForSingleObjectEx error: {:#x}", res);
+
         UINT32 written;
-        HRESULT hr{audio.mClient->GetCurrentPadding(&written)};
-        if(FAILED(hr))
+        uint len = 0;
+        HRESULT hr = S_FALSE;
+        if(sharemode == AUDCLNT_SHAREMODE_EXCLUSIVE)
         {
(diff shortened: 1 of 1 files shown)


alc/backends/wasapi.cpp
```diff
@@ -1232,28 +1232,49 @@ FORCE_ALIGN int WasapiPlayback::mixerProc()
...
-        if(FAILED(hr))
+        uint len = 0;
+        HRESULT hr = S_FALSE;
+        if(sharemode == AUDCLNT_SHAREMODE_EXCLUSIVE)
         {
-            ERR("Failed to get padding: {:#x}", as_unsigned(hr));
-            mDevice->handleDisconnect("Failed to retrieve buffer padding: {:#x}",
-                as_unsigned(hr));
-            break;
+            hr = audio.mClient->GetBufferSize(&written);
```
kcat, line 1251: Unless this value can change, it was already stored in `mOutBufferSize`, which is copied into the local variable `buffer_len`. It doesn't need to be queried each time. Also, since `written` is used to update `mPadding`, which is used for latency calculations, I'm not sure leaving it as the buffer size is a good idea (unless that is the actual amount of padding WASAPI holds for exclusive streams, i.e. the amount of audio still left to play before the new samples to be mixed).

ThreeDeeJay:
@kcat I'm still getting no audio on my Realtek card (which is probably what most PC users have) https://www.diffchecker.com/gR1s8cfX/
My Creative X-Fi USB works fine with the same settings, though: https://www.diffchecker.com/THjvIkax/
Here's a comparison of the logs of both cards using WASAPI exclusive: https://www.diffchecker.com/RG6XGpcW/
I noticed that the Realtek changes period_size from 144 to 160, and still doesn't work. I also get no audio if I manually set it to 160 or even values over 1K.

LAGonauta:
This should have been fixed in the reword.

ThreeDeeJay:
Can confirm 👌 
Realtek now works though it still post-resets to 160 even though I set 133 (anything lower crashes both), which sticks when using my Creative card: https://www.diffchecker.com/LCYzsRbR/

kcat:
> Can confirm 👌
> Realtek now works though it still post-resets to 160 even though I set 133 (anything lower crashes both), which sticks when using my Creative card: https://www.diffchecker.com/LCYzsRbR/

Possibly the Realtek device can't go lower than 160 (~3.333ms), while the Creative card can go a bit lower. If initializing fails with `AUDCLNT_E_BUFFER_SIZE_NOT_ALIGNED`, it gets the next supported period size, which would be 160. Values less than 133 are probably triggering a different initialization error (`AUDCLNT_E_BUFFER_SIZE_ERROR` maybe?), which causes a failure and the delayed crash.

ThreeDeeJay:
Replied in the main thread https://github.com/kcat/openal-soft/issues/682#issuecomment-2568690425

PR #1093, opened 2025-01-05
Use GHC::filesystem for std::filesystem compatibility

### LINKED ISSUE #1065: fatal error: 'filesystem' file not found
openal-soft 1.24.0 doesn't build on macOS 10.14 or earlier:

```
utils/makemhr/loaddef.cpp:33:10: fatal error: 'filesystem' file not found
#include <filesystem>
         ^~~~~~~~~~~~
```

Full build log: https://build.macports.org/builders/ports-10.14_x86_64-builder/builds/223612/steps/install-port/logs/stdio

It builds on macOS 12 and later.

This is a regression; 1.23.1 didn't have this problem.

`std::filesystem` is part of C++17 but the version of libc++ included with macOS 10.14 and earlier doesn't support it. Compiling with a newer version of clang that supports C++17 is not sufficient.

dpogue:
This brings in https://github.com/gulrak/filesystem as a "polyfill" for the C++17 filesystem API, particularly on older macOS versions where it's not part of the system C++ library. This should resolve #1065.

If possible, it will use the std::filesystem API, but when targeting macOS versions without support it will use the GHC implementation instead.

This uses the latest master branch of gulrak/filesystem because the current tagged release doesn't work properly on macOS.

I opted to just drop this in the common folder for ease of use, since it's a single file and that's what was done with pffft, but I could try to put it in a top-level vendor folder similar to fmt if that's the preferred approach.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -610,6 +610,8 @@ set(COMMON_OBJS
     common/comptr.h
     common/dynload.cpp
     common/dynload.h
+    common/filesystem.cpp
+    common/filesystem.h
     common/flexarray.h
     common/intrusive_ptr.h
     common/opthelpers.h
--- a/alc/alconfig.cpp
+++ b/alc/alconfig.cpp
@@ -34,7 +34,6 @@
 #include <array>
 #include <cctype>
 #include <cstdlib>
-#include <filesystem>
 #include <fstream>
 #include <istream>
 #include <limits>
@@ -47,6 +46,7 @@
 #include "alstring.h"
 #include "core/helpers.h"
 #include "core/logging.h"
+#include "filesystem.h"
 #include "strutils.h"
 
 #if ALSOFT_UWP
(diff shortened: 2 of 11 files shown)


kcat:
I'm not very keen on adding a custom implementation of an API that's part of the required standard, for an OS that's over 7 years old and didn't update to improve support. Simple workarounds would be fine, but a full implementation seems much. Especially since the next (non-point) release will be increasing the standard requirement to C++20, so won't be useful just after being added.

dpogue:
Agree that a full filesystem implementation is probably overkill, particularly given how minimal the use of it is in openal-soft, but this seemed like the option that involved the least code changes by swapping in something with the same API.

The issue on macOS is that the C++ runtime is provided by the system and doesn't get updated. So even if you build with a newer compiler version, you can't use APIs that weren't part of the system C++ runtime of the target system. It's theoretically possible to do static linking, but that's very uncommon and potentially introduces incompatibilities with system frameworks like CoreAudio. Previous attempts at mixing C++ runtimes have resulted in issues like pointers being malloc'ed with one runtime and free'd in another and throwing errors.

Unfortunately it comes down to implementation details which parts of the C++ standard are compiled away and which parts involve additions to the runtime, but filesystem is definitely one that isn't available on older macOS versions.

PR #1099, opened 2025-01-23
set cmake_minimum_required version to 3.26

2bt:
Generator expression BUILD_LOCAL_INTERFACE requires cmake minimum version 3.26

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # CMake build file list for OpenAL
 
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.26)
 enable_testing()
 
 if(APPLE)


petk:
When trying to find OpenAL with find_package(OpenAL CONFIG) the error is emitted on CMake 4.x versions.

Edit: I see, this change is for the root CMakeLists.txt. The error I'm mentioning is related to this:

https://github.com/kcat/openal-soft/blob/9d695bd3b603e8427b3a40f0c27708bf56a9ecac/OpenALConfig.cmake.in#L1

kcat:
> When trying to find OpenAL with find_package(OpenAL CONFIG) the error is emitted on CMake 4.x versions.
> 
> Edit: I see, this change is for the root CMakeLists.txt. The error I'm mentioning is related to this:
> 
> https://github.com/kcat/openal-soft/blob/9d695bd3b603e8427b3a40f0c27708bf56a9ecac/OpenALConfig.cmake.in#L1

What error are you getting?

petk:
This one (That CMake 3.1 should be increased in the config file):

```txt
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/OpenAL/OpenALConfig.cmake:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
Call Stack (most recent call first):
  CMakeLists.txt:20 (find_package)
```

kcat:
Please [create an issue](https://github.com/kcat/openal-soft/issues/new) about that, since it's not related to this PR.

petk:
Opened here https://github.com/kcat/openal-soft/issues/1161

PR #1131, opened 2025-04-29
fix redefinition warning

maxbachmann:
This fixes a redefinition warning when WIN32_LEAN_AND_MEAN is already defined by the toolchain.

This is the case when building for the xbox.


kcat:
I think it may be better to set this in CMakeLists.txt instead of individual files. If the OpenAL Soft is building with it defined "globally" by the toolchain you're using, then there shouldn't be any sources that depend on it not being defined, and defining it in CMakeLists.txt helps ensure it's not forgotten for some files.

maxbachmann:
I update it to set it from CMakeLists.txt

kcat (approved):

PR #1132, opened 2025-05-02
Port to Qt6

a17r:
Please take an extra long look over the QStringRef => QStringView change.


utils/alsoft-config/mainwindow.cpp
```diff
@@ -882,7 +882,7 @@ void MainWindow::loadConfig(const QString &fname)
             }
             else if(backend.size() > 1)
             {
-                QStringRef backendref{backend.rightRef(backend.size()-1)};
+                QStringView backendref = backend.right(backend.size()-1);
```
kcat, line 885: `right` seems to return a new `QString`, which would be assigning it to a non-owning QStringView and leave it dangling. I think this needs to be

```c++
auto backendref = QStringView{backend}.right(backend.size()-1);
```

kcat (changes requested):

utils/alsoft-config/mainwindow.cpp
a17r: thanks!

kcat (approved):

PR #1147, opened 2025-06-14
cmake: improve MINGW build

oltolm:
This is a small cleanup for MSYS2.
1. I removed `__USE_MINGW_ANSI_STDIO` because it's an internal define that should not be used.
1. Removed MINGW specific code that produces `OpenAL32.dll` and `OpenAL.dll.a`. Those files are produced properly.
1. Added support for CCache.
1. Fixed a warning in `common/win_main_utf8.h`.


CMakeLists.txt
```diff
@@ -1267,19 +1268,15 @@ if(ALSOFT_UPDATE_BUILD_VERSION AND GIT_FOUND AND EXISTS "${OpenAL_SOURCE_DIR}/.g
...
+            -DLIB_VERSION_NUM=${LIB_VERSION_NUM} -DSRC=${OpenAL_SOURCE_DIR}/version.h.in
+            -DDST=${OpenAL_BINARY_DIR}/version.h -P ${OpenAL_SOURCE_DIR}/version.cmake
         COMMAND ${CMAKE_COMMAND} -E touch "${OpenAL_BINARY_DIR}/version_witness.txt"
-        WORKING_DIRECTORY "${OpenAL_SOURCE_DIR}"
         MAIN_DEPENDENCY "${OpenAL_SOURCE_DIR}/version.h.in"
         DEPENDS "${GIT_DIR}/index" "${OpenAL_SOURCE_DIR}/version.cmake"
         VERBATIM
     )
-
-    add_custom_target(alsoft.build_version DEPENDS "${OpenAL_BINARY_DIR}/version_witness.txt")
```
kcat, line 1282: According to the cmake docs, this target is necessary when multiple other targets depend on the produced output file (directly or transitively): https://cmake.org/cmake/help/latest/command/add_custom_command.html#example-generating-files-for-multiple-targets

> If multiple independent targets need the same custom command output, it must be attached to a single custom target on which they all depend.

The "version_witness.txt" file is specified as the output to prevent CMake from thinking the command always needs to run and update version.h (along with anything that depends on it) when one of the command's dependencies updated but the output file doesn't need to be modified by the command. "version_witness.txt" is always updated when the dependencies updated (with the `touch` command to ensure its timestamp is updated) and won't run again until necessary, while version.h will only change as needed.

This in turn depends on using the `BYPRODUCTS` line for the Ninja generator to specify where the generated header is made, but not list it as a direct output requiring the command to run when its timestamp is older: https://cmake.org/cmake/help/latest/policy/CMP0058.html#policy:CMP0058

CMakeLists.txt
```diff
@@ -174,15 +180,9 @@ set(LINKER_FLAGS_RELEASE )
 set(EXTRA_LIBS )
 
 if(WIN32)
-    set(CPP_DEFS ${CPP_DEFS} _WIN32 NOMINMAX WIN32_LEAN_AND_MEAN)
-    if(MINGW)
-        set(CPP_DEFS ${CPP_DEFS} __USE_MINGW_ANSI_STDIO)
```
kcat, line 179: This needs to be set to make printf and other stdio functions use MinGW's standards-compliant versions instead of the ones from Windows' ancient msvcrt.dll. It is intended for users to set it.

CMakeLists.txt
```diff
@@ -3,6 +3,12 @@
 cmake_minimum_required(VERSION 3.13)
 enable_testing()
 
+find_program(CCACHE ccache)
+if(CCACHE)
+    set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
+    set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE})
+endif()
```
kcat, line 10: This seems like it should be put under an option, so it's not forced to be used just because ccache exists on the system.

kcat (changes requested):

CMakeLists.txt
oltolm: This was the case in the past. Now it is automatically set in mingw32 and mingw64 which use `msvcrt.dll`. It should not be set for all other [environments](https://www.msys2.org/docs/environments/) that use `ucrt`. I even asked this question in their discord and they confirmed it.

CMakeLists.txt
oltolm: I can do it.

CMakeLists.txt
```diff
@@ -1559,36 +1555,6 @@ target_compile_definitions(${IMPL_TARGET}
...
-            # Technically OpenAL32.def was created by the build, but cmake
-            # doesn't recognize it due to -Wl,--output-def,OpenAL32.def being
-            # manually specified. But declaring the file here allows it to be
-            # properly cleaned, e.g. during make clean.
-            BYPRODUCTS OpenAL32.def OpenAL32.lib
-            COMMENT "Stripping ordinals from OpenAL32.def and generating OpenAL32.lib..."
-            VERBATIM
-        )
-    endif()
-endif()
```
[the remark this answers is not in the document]
oltolm: In that case I will close the MR. I didn't know that.

Issue #1152, opened 2025-07-03
MacOS and iOS builds failing has prevented new releases for months


ThreeDeeJay:
I just noticed that because the MacOS and iOS builds have been failing, the OpenAL Soft Github latest pre-release hasn't been updated in *2 months*.

![Image](https://github.com/user-attachments/assets/c8171c80-4955-4ff4-82e9-4f75d105fca5)

![Image](https://github.com/user-attachments/assets/75c7194c-f90d-4a7e-9e8c-1166b93d9d7a)

So at the moment, it's 156 commits behind https://github.com/kcat/openal-soft/compare/eb7c7b7ff2a71ae703dc2228e1feb936a6aaa50b...78f62702080aed3c4c5935acd35d333b0074bb39

kcat:
Yes, unfortunately there's some problem with the macOS and iOS builds that I haven't been able to solve. The actual error is that it doesn't seem to recognize `std::views::join`, but the correct header is being included and other `std::views::*` stuff is working fine. Apple's own compliance page mentions the relevant C++20 features are supported with this compiler, and I can't find anything online about macOS missing or having problems with `std::views::join` specifically, so there's apparently something I'm missing or some easy/common workaround that others don't need explained.

kcat:
This has been addressed with commit da72e1929825916bf6322cb7a7fdd2edc439a8d2.

PR #1153, opened 2025-07-04
Always run release job, ignoring Apple build failures

ThreeDeeJay:
Simple workaround for https://github.com/kcat/openal-soft/issues/1152, so that Windows builds are released again even if MacOS and iOS builds fail.

image

https://github.com/ThreeDeeJay/openal-soft/releases/tag/latest

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,7 +191,7 @@ jobs:
       CommitDate: ${{env.CommitDate}}
 
   release:
-    if: github.event_name != 'pull_request'
+    if: always() && github.event_name != 'pull_request'
     needs: build
     runs-on: ubuntu-latest
     steps:


kcat:
Will this be a problem if there's a build problem that happens to slip into the other builds?

ThreeDeeJay:
Good question. I think if say, the Windows builds fail, the release upload steps would fail too because it wouldn't find the file.
So maybe I should instead check for Windows builds succeeding instead of always running.

ThreeDeeJay:
Hmm yeah I couldn't find a way to explicitly check for only both specific jobs in the matrix (Win32 and Win64) to succeed before running the release job. Apparently `needs: build` will always wait for all the jobs in the matrix to complete even if some fail.

But I still don't see how this could exacerbate the issue of jobs failing. 
IIUC they (and the run status overall) will still be flagged as failed, but with this PR if at least both Win32 and Win64 succeed, the release job would still move forward with releasing those binaries. Otherwise, commands that rely on those 2 builds to succeed with a generated DLL existing (like `cp "build/soft_oal-Win32-Release/soft_oal.dll"` and `cp "build/soft_oal-Win64-Release/soft_oal.dll"`) would fail and then I presume the release job would also fail, without creating a new release.

kcat:
I worked around the issue for now by using macos-15 on the CI builds instead of macos-latest (14), which seems to build it successfully.

ThreeDeeJay:
Perfect, then I guess this won't be needed after all 👌

Issue #1165, opened 2025-08-20
ffmpeg automatic dependency breaks the build if SDL3 happens to be installed


smcv:
openal 1.24.2.

For context, Debian normally builds openal-soft in a minimal environment with only the required packages. In the configuration used by Debian, features that require SDL3, libsndfile and ffmpeg are not enabled, and development headers for those packages are not installed. It's easiest to demonstrate this by building Debian's packaged version of openal-soft 1.24.2 rather than the upstream source:

```
$ podman run --rm -it debian:forky-slim
# sed -i -e 's/^Types: deb/& deb-src/' /etc/apt/sources.list.d/debian.sources
# apt update
# apt full-upgrade
# apt build-dep openal-soft
# apt source openal-soft
# cd openal-soft-*
# dpkg-buildpackage
...
make[2]: Entering directory '/openal-soft-1.24.2'
dh_auto_configure -- -DCMAKE_VERBOSE_MAKEFILE=ON -DLIB_SUFFIX="/x86_64-linux-gnu" -DEXAMPLES=OFF ..
	cd build-tree && DEB_PYTHON_INSTALL_LAYOUT=deb PKG_CONFIG=/usr/bin/pkg-config cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DFETCHCONTENT_FULLY_DISCONNECTED=ON -DCMAKE_INSTALL_RUNSTATEDIR=/run -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=ON "-GUnix Makefiles" -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DCMAKE_VERBOSE_MAKEFILE=ON -DLIB_SUFFIX=/x86_64-linux-gnu -DEXAMPLES=OFF .. ..
... and the build succeeds
```

But if I add SDL3 development files and repeat the build, it fails at the configure stage:

```
# apt install libsdl3-dev
# dpkg-buildpackage
...
[...]

PR #1166, opened 2025-08-20
Reduce automagic dependencies

smcv:
* build: Add an option for whether to use libsndfile
    
    OS distributions typically want to build a package with a predictable
    set of options enabled, even if they are building in a non-minimal
    environment where additional development headers/libraries are present.
    For example,
    <https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies>
    in Gentoo documentation explains this topic well.
    
    With `-DALSOFT_SNDFILE=ON` (the default), libsndfile will still be used
    if present, or not used if absent. With `-DALSOFT_SNDFILE=OFF`, it will
    not be used, even if it happens to be present.

* build: Add an option for whether to use ffmpeg
    
    Same reasoning as for the previous commit.

* build: Add an option for whether to use SDL3
    
    Same reasoning as for the previous commit.

Helps: #1165


smcv:
I think this is still worth having even after eb2b8b8e "Check if the header exists before reading from it" and 67b1a3aa "Don't check for ffmpeg if ALSOFT_EXAMPLES is false" fixed #1165, because it allows a user/packager to build with examples enabled, but without all of the examples' optional dependencies.

kcat:
CMake has [a built-in feature to disable optional package dependencies](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html). Setting `-DCMAKE_DISABLE_FIND_PACKAGE_SndFile=TRUE` will make `find_package(SndFile)` fail to find the SndFile package for example, even if it's available on the system, and the script will continue without it. Similar for `SDL3`, `FFmpeg`, and others.

smcv:
Thanks, I wasn't aware of that. This PR is probably unnecessary, then.

PR #1183, opened 2025-09-28
Add an XML registry

Perksey:
_As discussed in email with @kcat, most of this PR description is copied from that email exchange_

---

# Background 
I’m a contributor of the open-source Silk.NET bindings library for C#, and OpenAL is one of the bindings we provide along with OpenGL.

Traditionally, our OpenAL bindings have been handwritten, whereas our OpenGL bindings have been generated based on its XML registry that defines the APIs, the extensions they belong to, etc.
As a result, we’ve often lagged behind the capabilities of OpenAL Soft as there’s a lot of steps between OpenAL Soft getting a new feature and it being added to our bindings. I’d imagine a great many of the bindings libraries in other languages probably struggle with similar issues.

In Silk.NET 3.0, we’ve evaluated how we can move our bindings to evolve our bindings at a similar rate to OpenAL Soft itself. We’ve recently made a lot of progress with a Clang-based bindings generator, so we can ingest OpenAL Soft headers directly.
However, this does not include the other manual bits that we’ve effectively “invented” like the extension groupings and the grouping of ALenum #defines to strong types.

# Contribution
This PR upstreams an XML registry similar to OpenGL's XML registry. Effectively, the workflow would be that this XML would be a single source of truth, and e.g. the AL headers would be generated from this XML. This obviously has workflow implications for OpenAL Soft contributors, which is why I've put a lot of effort in to ensure that this isn’t overly burdensome and hopefully generates some value for you as well as the downstream consumers of bindings such as Silk.NET.

If the existing OpenGL way of doing things doesn’t work for you, we’re happy to work with you to figure out something that works for us both. We’d of course write any scripts and contribute them into the repo to make this workflow work - I’m well aware that this at face value seems like a lot of extra stuff that is unreasonable if it’s at all burdensome to you or your contributors.

Please let me know what you think!

# Future Work

Today, the headers are generated by a very crude python script I wrote. In the future, we probably want tooling that is more aligned to what Khronos has (as these tools validate things like dependency chains, includes, etc; whereas today we have a lot of magic in the scripts e.g. the preambles, which would not pass their tools' validation.

The main motivator in the python script being the way it is is to ensure that the changes to the headers are minimal (ideally only whitespace or moving things around), and I believe I've accomplished this. If we're more receptive to a bigger break in the future (not API/ABI breaking, but a big change in how the headers work and/or this workflow) then we can evaluate it then.


Perksey:
Btw, it helps to have "hide whitespace" when reviewing this PR.

include/AL/alc.h
```diff
@@ -148,156 +148,191 @@ typedef void ALCvoid;
...
- * Device's extended specifier string.
- *
- * If device handle is NULL, it is instead a null-character separated list of
- * strings of known extended device specifiers (list ends with an empty string).
- */
-#define ALC_ALL_DEVICES_SPECIFIER                0x1013
-
-
-#ifndef ALC_NO_PROTOTYPES
+#ifndef AL_NO_PROTOTYPES
```
kcat, line 177: Not sure what's causing this, but this is a breaking change, hiding the standard ALC functions with `AL_NO_PROTOTYPES` instead of `ALC_NO_PROTOTYPES`.

registry/xml/al.xml
```diff
@@ -0,0 +1,5469 @@
...
+                <enum name="AL_BANDPASS_MIN_GAINHF"/>
+                <enum name="AL_BANDPASS_MAX_GAINHF"/>
+                <enum name="AL_BANDPASS_DEFAULT_GAINHF"/>
+            </require>
+            <require api="al">
+                <enum name="AL_BANDPASS_MIN_GAINLF"/>
+                <enum name="AL_BANDPASS_MAX_GAINLF"/>
+                <enum name="AL_BANDPASS_DEFAULT_GAINLF"/>
+            </require>
+            <r
```
kcat, line 4815: Typo, should be `AL_SOFT_buffer_samples`.

registry/xml/al.xml
```diff
@@ -0,0 +1,5469 @@
...
+        <command export="alc">
+            <proto>void <name>alcDevicePauseSOFT</name></proto>
+            <param><ptype>ALCdevice</ptype> *<name>device</name></param>
+        </command>
+        <command export="alc">
+            <proto>void <name>alcDeviceResumeSOFT</name></proto>
+            <param><ptype>ALCdevice</ptype> *<name>device</name></param>
+        </command>
+        <command export="alc">
+            <proto>const <ptype>ALchar</ptype>* <name>alcGetStringiSOFT</name></proto>
```
kcat, line 3468: Should be `ALCchar` instead of `ALchar`.

kcat (changes requested):

registry/xml/al.xml
kcat: Hmm, not sure what's happening with the line references. It should be 4815, not 4241...4244.

include/AL/alext.h
```diff
@@ -626,38 +621,35 @@ AL_API void AL_APIENTRY alGetBufferPtrvSOFT(ALuint buffer, ALenum param, ALvoid
 #endif
 
 #ifndef ALC_SOFT_output_mode
-#define ALC_SOFT_output_mode
+#define ALC_SOFT_output_mode 1
 #define ALC_OUTPUT_MODE_SOFT                     0x19AC
 #define ALC_ANY_SOFT                             0x19AD
-/*#define ALC_MONO_SOFT                            0x1500*/
-/*#define ALC_STEREO_SOFT                          0x1501*/
```
kcat, line 633: Is it possible to keep the duplicate defines commented out like these? For people that look in the header can see they're part of the macro list, but can't be defined since they're already defined earlier (and for some other extensions, to see they're not empty like `AL_SOFT_source_length`).

Perksey: Of course! Added back.

registry/xml/al.xml
Perksey: Fixed, thanks.

include/AL/alext.h
```diff
@@ -569,34 +579,32 @@ AL_API void AL_APIENTRY alGetPointervSOFT(ALenum pname, void **values) AL_API_NO
...
+typedef void (AL_APIENTRY *LPALGETBUFFERPTRVSOFT)(ALuint buffer, ALenum param, ALvoid **ptr) AL_API_NOEXCEPT17;
 #ifdef AL_ALEXT_PROTOTYPES
-AL_API void AL_APIENTRY alBufferCallbackSOFT(ALuint buffer, ALenum format, ALsizei freq, ALBUFFERCALLBACKTYPESOFT callback, ALvoid *userptr) AL_API_NOEXCEPT;
-AL_API void AL_APIENTRY alGetBufferPtrSOFT(ALuint buffer, ALenum param, ALvoid **ptr) AL_API_NOEXCEPT;
-AL_API void AL_APIENTRY alGetBuffer3PtrSOFT(ALuint buffer, ALenum param, ALvoid **ptr0, ALvoid **ptr1, ALvoid **ptr2) AL_API_NOEXCEPT;
-AL_API void AL_APIENTRY alGetBufferPtrvSOFT(ALuint buffer, ALenum param, ALvoid **ptr) AL_API_NOEXCEPT;
+void AL_APIENTRY alBufferCallbackSOFT(ALuint buffer, ALenum format, ALsizei freq, ALBUFFERCALLBACKTYPESOFT callback, ALvoid *userptr) AL_API_NOEXCEPT;
+void AL_APIENTRY alGetBufferPtrSOFT(ALuint buffer, ALenum param, ALvoid **ptr) AL_API_NOEXCEPT;
+void AL_APIENTRY alGetBuffer3PtrSOFT(ALuint buffer, ALenum param, ALvoid **ptr0, ALvoid **ptr1, ALvoid **ptr2) AL_API_NOEXCEPT;
+void AL_APIENTRY alGetBufferPtrvSOFT(ALuint buffer, ALenum param, ALvoid **ptr) AL_API_NOEXCEPT;
```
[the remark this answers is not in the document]
Perksey: Damn and I thought I had attention to detail. Fixed :)

Perksey:
Already successfully generating bindings using this new al.xml btw :tada: https://github.com/dotnet/Silk.NET/pull/2494/

kcat:
Looks like builds are failing because EFX isn't coming through with `alext.h`. Which seems to be because `efx.h` includes a `ALC_EXT_EFX` macro guard, which is also used in `alext.h` around the `#include "efx.h"`. It should probably be removed from the latter if the former will have it.

kcat:
Also, I'm curious how this could be used to generate [the C++ modules](https://github.com/kcat/openal-soft/tree/master/modules), `al.cppm`, `alc.cppm`, `alext.cppm`, and `efx.cppm`. The primary differences are:

* No header guards, since they're source files and not header files. Instead they have `module;` and `export module openal.<name>;` statements at/near the top, with any extra `#includes` being between them.
* C++ trailing return types (e.g. `AL_API auto AL_APIENTRY alIsSource(ALuint source) AL_API_NOEXCEPT -> ALboolean;` instead of `AL_API ALboolean AL_APIENTRY alIsSource(ALuint source) AL_API_NOEXCEPT;`).
* C++ `using` instead of `typedef` (e.g. `using ALchar = char;` instead of `typedef char ALchar;`, or for function pointer types, `using LPALISSOURCE = auto (AL_APIENTRY*)(ALuint source) AL_API_NOEXCEPT -> ALboolean;` instead of `typedef ALboolean (AL_APIENTRY *LPALISSOURCE)(ALuint source) AL_API_NOEXCEPT17;`).
* Enums are defined as global constexpr variables instead of macros (since macros aren't exported from modules).
* Transitive exports, like `alext.h` including `efx.h` to purposely expose EFX, is done by `export import openal.efx;`.
* Macros likes `AL_VERSION_1_1` or guards like `ALC_EXT_EFX` are unneeded since they aren't exported from the module so user code can't test them with the preprocessor, and they're unnecessary for building the module (they don't hurt to be there, but they're not useful to have).

I'm not asking for you to do this, unless you want to. But I'm not very familiar with Python, so what would I need to know to be able to add this to the Python script?

Perksey:
> I'm not asking for you to do this, unless you want to. But I'm not very familiar with Python, so what would I need to know to be able to add this to the Python script?

If this makes it easier for you, I'm happy to do this. Ultimately Silk.NET has a vested interest in this XML registry being maintained, and if it proves painful for you as a maintainer then you're naturally going to want to cast it aside. 

~~However, if you're going to be adding C++ specific things like a C++ module then there's likely a larger design discussion to be had. For instance, `vulkan.cppm` is part of the `Vulkan-Hpp` project which uses the XML registry to generate more C++-ified bindings (e.g. `enum class`es, `namespace`d APIs, etc). This is done using another set of bespoke scripts which coexist with the C header generation scripts.~~ EDIT: didn't realise the C++ modules had already been added. Will look into writing this up now.

Ultimately, the current Python script is very dumb and even as a non-Python programmer you'll probably be able to figure out what's going on. Admittedly if I knew the audience wasn't as familiar as Python, I would've held back in some places (e.g. we're using generator expressions, which are notoriously a rather difficult concept to grasp) but most of the code there is just fluff to try and get the new headers to match the old headers as much as possible (e.g. 80 of the 623 lines are just meticulously formatting documentation strings) - there is a lot that could be simplified if we were more breaking.

Perksey:
> Looks like builds are failing because EFX isn't coming through with alext.h. Which seems to be because efx.h includes a ALC_EXT_EFX macro guard, which is also used in alext.h around the #include "efx.h". It should probably be removed from the latter if the former will have it.

Yep, just came across this when generating the bindings as well - none of the EFX APIs popped through! I've removed this extraneous guard in my latest commit.

kcat:
The EFX functions seem to be missing `AL_API` too (sorry, didn't notice this earlier with the others).

> EDIT: didn't realise the C++ modules had already been added. Will look into writing this up now.

Yeah, the C++ modules are just a "simple" conversion of the headers to be as functionally equivalent as possible. Ideally, someone could just swap out the include headers for import modules, and most existing OpenAL code would work as-is. A more proper C++ interface would need additional work (and OpenAL extensions), maybe as part of an API refresh or something.

> most of the code there is just fluff to try and get the new headers to match the old headers as much as possible (e.g. 80 of the 623 lines are just meticulously formatting documentation strings) - there is a lot that could be simplified if we were more breaking.

As long as its readable and understandable to someone perusing the headers, and functionally the same to the compiler and code using it, I don't think it matters too much if things get changed up a bit.

One other thing to bring up, I think it may be worthwhile to add a small license blurb to the generated headers. Like how the modules have the Unlicense text in them to distinguish them from being LGPL like the main library, allowing people to copy them into their project and make any changes they need to use them without worrying about licensing (extra important for modules since they need to be part of the target app's build; even though APIs can't be copyrighted, I could see including the header files themselves and/or compiling the module files in their own project as a potential concern).

Perksey:
Fixed the AL_API for the EFX headers, and I've also made the modules generate from the XML as well. I've split the Python scripts in two parts: reg.py and genheaders.py. The first parses the XML, and the latter generates the headers and modules (I considered splitting these into two separate scripts but there'd be a lot of duplicated code). I've also added an unlicense file header to the headers as well.

Hopefully this is everything! Let me know if there's anything I've missed.

registry/scripts/genheaders.py
```diff
@@ -0,0 +1,801 @@
...
+#ifndef AL_API
+ #define AL_API ALC_API
+#endif
+#define AL_APIENTRY ALC_APIENTRY
+#define AL_API_NOEXCEPT ALC_API_NOEXCEPT
+
+#define ENUMDCL inline constexpr auto
+
+typedef std::int64_t _alsoft_int64_t;
+typedef std::uint64_t _alsoft_uint64_t;
```
kcat, line 291: These need to be moved to under the `export module openal.ext` line. Only preprocessor directives (and comments) can appear between the `module;` and `export module ...` lines.

kcat (approved):

PR #1189, opened 2025-10-10
Record dynamic dependencies in .note.dlopen ELF section

### LINKED ISSUE #1188: Record dynamic dependencies in `.note.dlopen` ELF section
Hi there,

openal-soft attempts to load quite some libraries at run-time. However, it is currently impossible by inspecting the `libopenal.so` library to tell (a) which other libraries are loaded dynamically and (b) which dynamic dependencies are more important than others. This makes it hard for e.g. Linux distributions to declare the best set of dependencies on other libraries when packaging openal-soft.

To mitigate this, there is a new tool available in Debian [1] that inspects some additional information hidden in the `.note.dlopen` section of ELF files. However, adding this information is still due to the original library itself. The spec is available at [2].

As one of the first projects, the SDL team has adopted this workflow and adds priority information to its dynamic dependencies [3], which the Debian package can extract and translate into package dependencies.

Do you think it would make sense to adopt his to openal-soft as well (I do) and would you like me to propose a PR based on the original implementation found in SDL?

Cheers,

 - Fabian

[1] https://manpages.debian.org/testing/dh-dlopenlibdeps/dh_dlopenlibdeps.1.en.html
[2] https://systemd.io/PACKAGE_METADATA_FOR_EXECUTABLE_FILES/
[3] https://github.com/libsdl-org/SDL/blob/HEAD/src/dynapi/SDL_dynapi_dlopennote.h

fabiangreffrath:
Code is mostly taken from SDL3

Fixes #1188

### DIFF
--- a/alc/backends/alsa.cpp
+++ b/alc/backends/alsa.cpp
@@ -1147,12 +1147,14 @@ auto AlsaCapture::getClockLatency() -> ClockLatency
 
 #define ALSA_LIB "libasound.so.2"
 
+#if HAVE_DYNLOAD
 OAL_ELF_NOTE_DLOPEN(
     "backend-alsa",
     "Support for the ALSA backend",
     OAL_ELF_NOTE_DLOPEN_PRIORITY_SUGGESTED,
     ALSA_LIB
 );
+#endif
 
 auto AlsaBackendFactory::init() -> bool
 {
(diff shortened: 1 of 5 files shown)


fabiangreffrath:
~Dammit, I accidently killed the `build/.empty` file and now GIT doesn't let me add it back, because, well, it's empty.~

Never mind.

Issue #1191, opened 2025-10-22
[Android] No audio when upgrading libopenal.so


ThreeDeeJay:
Recently, I made a [Magisk module](https://github.com/ThreeDeeJay/OpenALSoftGlobalHRTF) to enable OpenAL Soft's HRTF globally in Android since most apps/games lack an HRTF toggle, so naturally I decided to test it. So using a [root-capable file manager](https://mixplorer.com/beta/) I searched `/data/app/` for `openal` to find apps using the library and found [a few hits](https://github.com/user-attachments/assets/8cefe250-ab91-4d9d-8dfd-ab9df92a816d). And as expected, I did get full 3D audio in most, but some used an ancient build of OpenAL Soft which didn't seem to play nice with my MHR so I decided to upgrade them to the [latest](https://nightly.link/kcat/openal-soft/workflows/ci/master/soft_oal-Android_arm64-v8a-Release.zip). But weirdly enough, I was getting no audio, so I got the logs:

- [OpenMW](https://gitlab.com/cavebros/openmw-android-docker/-/releases) (Morrowind port): https://www.diffchecker.com/JrFJokMW/
  - I noticed it's already using a recent build, but maybe it doesn't like being forced to use OpenSL instead of Oboe?. 
- [idTech4A++](https://github.com/glKarin/com.n0n3m4.diii4a)  ([multi-Doom/Quake/etc](https://github.com/glKarin/com.n0n3m4.diii4a/blob/2da6518a3bde4a7af2169497705f2188e66b8cbc/Q3E/src/main/jni/CMakeLists.txt#L28-L52) port) https://www.diffchecker.com/ditBXKvv/
- [RVGL](https://rvgl.org/downloads/rvgl_23.1030a1_android.apk) (Re-Volt port): https://www.diffchecker.com/QodliKdI/

I got the errors reported in https://github.com/kcat/openal-soft/issues/1139 in all of them after upgrading.
Though if I had to take a shot in the dark, I'd guess OpenMW and idTech4A++ don't play nice when switching from Oboe to OpenSL, though RVGL was already working with it so it could be something else entirely, especially since it was using an ancient build (1.19.1) 🤔

kcat:
> Though if I had to take a shot in the dark, I'd guess OpenMW and idTech4A++ don't play nice when switching from Oboe to OpenSL

[...]

PR #1192, opened 2025-10-23
Enable Oboe in Android CI builds

ThreeDeeJay:
Fixes https://github.com/kcat/openal-soft/issues/1191 and potentially [other issues](https://github.com/search?q=repo%3Akcat%2Fopenal-soft%20oboe&type=issues) people have been running into.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
 
     - name: Setup Oboe for Android
       if: ${{ contains(matrix.config.name, 'Android') }}
-      run: git clone https://github.com/google/oboe
+      run: git clone https://github.com/google/oboe --depth=1 --branch 1.10.0
 
     - name: Get current commit tag, short hash, count and date
       run: |


kcat:
It's probably good to enable the Oboe backend for the CI, but this doesn't really fix #1191 since that issue is looking like a problem with the OpenSL backend. This is just side-stepping the issue by not using the OpenSL backend, so whatever the problem is is still there.

.github/workflows/ci.yml
```diff
@@ -121,14 +123,20 @@ jobs:
...
             build_type: "Release"
           }
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: '0'
 
+    - name: Setup Oboe for Android
+      if: ${{ contains(matrix.config.name, 'Android') }}
+      run: git clone https://github.com/google/oboe
```
kcat: Would probably be good to add `--depth=1` to not pull the whole history. Maybe also have it grab the latest release if possible, instead of the head.

ThreeDeeJay:
> It's probably good to enable the Oboe backend for the CI, but this doesn't really fix #1191 since that issue is looking like a problem with the OpenSL backend. This is just side-stepping the issue by not using the OpenSL backend, so whatever the problem is is still there.

Fair enough. 👌

PR #1223, opened 2026-01-08
Fix apple framework bundle ID not set

halx99:
Fix warning when build as Apple framework: warning: User-supplied CFBundleIdentifier value 'org.openal-soft.openal' in the Info.plist must be the same as the PRODUCT_BUNDLE_IDENTIFIER build setting value ''.


halx99:
refer: https://github.com/axmolengine/axmol/actions/runs/20805927830/job/59768781235#step:4:2276

PR #1256, opened 2026-03-22
Allow disabling dynamic loading of optional libs on Windows

solbjorn:
Currently, setting `ALSOFT_DLOPEN` to `OFF` on Linux disables dynamic library loading and instead requires them to be linked directly. However, on Windows the dynamic loading is always enabled, which might not always be wanted.

Set `HAVE_DYNLIB` to 0 on Windows as well if `ALSOFT_DLOPEN` is disabled and do not compile the corresponding code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -662,12 +662,17 @@ if(HAVE_LIBRT)
 endif()
 
 # Check for the dlopen API (for dynamically loading backend libs)
+set(HAVE_DYNLOAD 0)
 if(ALSOFT_DLOPEN)
     check_include_file(dlfcn.h HAVE_DLFCN_H)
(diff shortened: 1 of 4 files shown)


kcat:
Do note that dynamic loading won't be totally disabled on Windows this way, only backends that use these functions. The WASAPI backend uses COM, which may also dynamically load libraries. Currently DSound uses these functions to load `dsound.dll`, but it is possible to use COM for that to.

There's nothing wrong with this change, so I'm not opposed to it. Just be aware it won't have as much of an effect for the main Windows backends.

solbjorn:
Yes, sure. I created this due to that I have a monolithic exe linked statically and when I added PortAudio and enabled it for openal-soft, there was no option to still link it statically.

kcat (approved):

PR #1269, opened 2026-04-19
Expose function to setting config values

Raltyro:


kcat:
There's more needed before something like this can happen. If the function can be called by different threads, there needs to be a lock to ensure a thread doesn't try to write to the vector while another is reading or writing to it also.

There's also no guarantees about what options are valid (config options get added, deprecated, and removed with newer versions), the format of the option values (e.g. the `hq-mode` option in the `[decoder]` section being defined by the `decoder/hq-mode` key in an implementation detail, or that the `cf_level` option in the `[general]` section is `cf_level` instead of `general/cf_level`), or when they're used (some config options are only read once when the library initializes itself where changing it afterward will have no effect, some are re-read whenever a device is opened, some are re-read when the device is reconfigured (without the setting being used to determine if the device needs reconfiguring)).

Raltyro:
Maybe after thinking more about it, this way of being able to modify the default configurations is not viable at all, maybe i'll look more into it later, probably a way to have the configurations as a environment variables in ALSOFT_CONF?

PR #1276, opened 2026-05-05
Use find_program to find git executable for consistency

halx99:
It's useful when find host program only when cross-compiling


kcat:
I'm not sure I understand what this is fixing. The FindGit module seems to use `find_program` under the hood, but also checks locations and alternate names to improve detection on Windows, instead of a plain default search for a `git` executable. What's the problem this is trying to resolve?

PR #1294, opened 2026-06-10
Add RISC-V target detection

carlosqwqqwq:
## Why

openal-soft did not expose an explicit RISC-V target path in its CMake SIMD detection, which risked host-feature leakage into simulated or cross-target riscv configurations. This patch makes the generic RISC-V path explicit while keeping the existing scalar mixer path.

## What changed

- Detect RISC-V targets in the CMake architecture probe using `CMAKE_SYSTEM_PROCESSOR` and `CMAKE_C_COMPILER_TARGET`.
- Generate `ALSOFT_TARGET_RISCV` in `config_simd.h` for RISC-V configurations.
- Keep RISC-V builds on the existing default scalar/generic path.
- Prevent simulated `riscv64` configurations from enabling SSE, SSE2, SSE3, SSE4.1, or NEON through host header leakage.

## Verification

- Ran native CMake configuration with Ninja, MinGW GCC, static library output, and non-essential programs disabled.
- Built the native static library successfully with `cmake --build build-codex-native --target OpenAL --config Release -j 4`.
- Ran simulated `riscv64` Linux CMake configuration with `CMAKE_SYSTEM_PROCESSOR=riscv64` and `CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY`.
- Built the simulated `riscv64` static library successfully with `cmake --build build-codex-riscv --target OpenAL --config Release -j 4`.
- Checked generated `build-codex-riscv/config_simd.h`: `ALSOFT_TARGET_RISCV=1`, `HAVE_SSE=0`, `HAVE_SSE2=0`, `HAVE_SSE3=0`, `HAVE_SSE4_1=0`, `HAVE_SSE_INTRINSICS=0`, and `HAVE_NEON=0`.
- Checked `build-codex-riscv/build.ninja` and `compile_commands.json`: no `mixer_sse*.cpp`, `mixer_neon.cpp`, `-msse*`, `-mfpmath=sse`, or `arm_neon.h` references were present.

## Notes

This is a conservative portability patch. It does not add RVV mixers or RISC-V-specific DSP optimizations. No real riscv64 cross-toolchain or hardware validation was available on the local Windows host.


kcat:
If there's an issue with host header leakage, those should be fixed somehow. If some headers and functions are being detected from the host system (e.g. x86-64) that don't apply to the build target (e.g. RISC-V), that would cause more problems than SSE and NEON being enabled. Forcing SSE/NEON off when RISC-V is detected is a pretty bad hack, since it would prevent them from working even if they would work (via an API compatibility layer or something, for example).

I don't think setting `CMAKE_SYSTEM_PROCESSOR` to `riscv64` is enough to do a proper cross-compile. You'll need a proper cross-compiler that has the appropriate headers and sets the appropriate compile settings to build RISC-V binaries, and a CMake toolchain file that defines the proper system and compiler variables for the configuration. If the RISC-V compiler doesn't support SSE or NEON, they shouldn't be detected as supported.

carlosqwqqwq:
Thank you for the review, and sorry for the earlier hacky approach. In the new commit be9b3b06c (Fix SIMD detection for RISC-V cross-compiles), I removed the ALSOFT_TARGET_RISCV target-name blacklist entirely and replaced the SSE/SSE2/SSE3/SSE4.1/NEON checks with target-aware check_cxx_source_compiles(...) probes, so SIMD support is now detected from the actual target compiler instead of host header leakage or architecture-name heuristics. I also verified this with a real local riscv64-linux-gnu cross-build in Docker: the build falls back to the default scalar path, config_simd.h has HAVE_SSE* = 0 and HAVE_NEON = 0, no mixer_sse* or mixer_neon sources are included, and the resulting library is a RISC-V ELF shared object. Sorry again for the noise in the previous revision, and thank you for pointing me to the correct direction.

PR #1297, opened 2026-06-16
Fix compatibility with fmt 12.2.0

xiaozhuai:
## Problem

fmtlib/fmt deprecated `<fmt/core.h>` in commit [0007426](https://github.com/fmtlib/fmt/commit/0007426c2d8d22df6e56d3b4d109415242e683e0). Starting from fmt 12.2.0, using `<fmt/core.h>` causes compilation errors because the deprecated include is now opt-in only.

## Fix

Replace all `#include <fmt/core.h>` with `#include <fmt/format.h>`, which is the standard header in fmt 12.2.0+.

### DIFF
--- a/alc/alc.cpp
+++ b/alc/alc.cpp
@@ -101,7 +101,7 @@
 #include "effects/base.h"
 #include "export_list.h"
 #include "flexarray.h"
-#include "fmt/core.h"
+#include "fmt/format.h"
 #include "fmt/ranges.h"
 #include "gsl/gsl"
 #include "inprogext.h"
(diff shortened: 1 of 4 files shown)


kcat (approved):

solbjorn:
Out of curiosity: this project ships a copy of fmt 11.2.0 and doesn't use external fmt, what is this PR about?

kcat:
It will reduce the amount of changes needed in the main codebase to update the internal copy, since it's something that would need to be done anyway.

solbjorn:
Have you maybe considered a CMake option to use external fmt? I link with openal-soft statically and I also use upstream fmt in the project, which means I have 2 copies of it inside the exe :s

